Issue #175, opened 2026-01-09
Support grayscale (ZeroIsBlack)


mdsumner:
See #172 and so we can  point at gebco and obvious range scaling

https://source.coop/alexgleith/gebco-2024/GEBCO_2024.tif

kylebarron:
Yes, this will be pretty simple. Just needs an extra case here:
https://github.com/developmentseed/deck.gl-raster/blob/807be9ea7dca373780f558f96c5bcfe0243feba6/packages/deck.gl-geotiff/src/geotiff/render-pipeline.ts#L222-L225

We can create another module like the above cmyk/ycbcr/etc modules for converting the grayscale single-band to 3-band rgb grayscale

mdsumner:
ah, I had to warm into this a bit again ... with a full-globe longlat tif I see tile issues, with an Australian Albers image it works fine, butwith a polar (IBCSO) subset image it doesn't recognize the specific EPSG 9354 (... because *proj4js* doesn't ship with proj.db support - hence https://github.com/developmentseed/deck.gl-raster/issues/185 ). 

I'm not super au fait with the the workflow but I was able to merge those 3 PRs and build and host it but having non-definitive feeback still  - a thing that would help me most rn is the ability to input a COG url 

```
const params = new URLSearchParams(window.location.search);
const COG_URL = params.get('cog') || 'https://projects.pawsey.org.au/image-cogs/images/IBCSO_v2_digital_chart_bl.tif';
```

Would you like to convert cog-basic to allow that?   Maybe keep undocumented, but harmless so we can test without duplicating much. 

I will continue to explore and create test COGs and share a meaningful set.   But, in terms of a test fixture the GDAL approach is something like: 

```
 gdal pipeline ! read  /vsicurl/https://data.source.coop/alexgleith/gebco-2024/GEBCO_2024.tif ! clip --bbox -180,-85,180,85 ! resize --size 512,256 ! write --of COG --co COMPRESS=ZSTD --co OVERVIEW_COUNT=3  /tmp/mycogs/gebco_boxmerc_512.tif 
```

(the default overview calc won't create any because the raster is so small) when it's considered workable replace /tmp/<prefix> with the '/vsis3/mycogs/<prefix>' path and the env vars for auth/config opts ). It's somehwat easier to use 'reproject' (or gdalwarp) because you don't have to pipe line combos of output properties - crs/bbox/size/resolution - they are all optional and specifiable for a one step convert.  I'll keep notes and share.

PR #179, opened 2026-01-12
feat: Support grayscale photometric interpretation

kylebarron:
Closes https://github.com/developmentseed/deck.gl-raster/issues/175

### Change list

- Create new BlackIsZero and WhiteIsZero GPU shader modules
- Ensure that typed array type matches what the GPU expects for the given texture format. This had been preventing uint16 images from rendering

`uint8_1band_deflate_block128_unaligned_mask.tif` from https://github.com/developmentseed/geotiff-test-data/pull/50

image

And with uint16 1-band data https://github.com/developmentseed/geotiff-test-data/blob/7fa07ac10f9d54b2ac2e557117432cf8f2f80eef/rasterio_generated/fixtures/uint16_1band_scale_offset_info.md#L4

image

### DIFF
--- a/packages/deck.gl-geotiff/src/geotiff/render-pipeline.ts
+++ b/packages/deck.gl-geotiff/src/geotiff/render-pipeline.ts
@@ -11,7 +11,11 @@ import {
   WhiteIsZero,
 } from "@developmentseed/deck.gl-raster/gpu-modules";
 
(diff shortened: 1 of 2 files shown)


kylebarron:
> The functions `black_zero_to_rgb(float value)` and `white_zero_to_rgb(float value)` expect a float, but color is a vec4.

Yes, I might not have pushed my last commit on this branch

PR #180, opened 2026-01-13
feat: Handle signed integer data types

kylebarron:
## What I am changing

-

## How I did it

-

## How you can test it

-

## Related Issues

-


kylebarron:
This is not strictly true. Integer textures can use unorm samplers.

thatbudakguy:
Any plans to merge this? Would be nice to have support for this type of data.

kylebarron:
Yes, I'd like to merge something like this, but we might need an overhaul of how our "automatically infer the correct rendering given generic input" works. This PR itself is too stale and has too many conflicts to be merged directly.

PR #207, opened 2026-01-27
chore!: Bump to geotiff.js 3.0

kylebarron:
Waiting on upstream fixes: https://github.com/geotiffjs/geotiff.js/issues/512, https://github.com/geotiffjs/geotiff.js/issues/513


kylebarron:
We're switching from geotiff.js to `@cogeotiff/core`

Issue #185, opened 2026-01-14
proj4js *crs registration* workaround


mdsumner:
proj4js does not ship with `proj.db` and so uses workarounds for "common" codes like "EPSG:3031" by *registering* them. This obviously does not scale well to the available COGs out there. One way to work around this is to get the required OGCWKT text from spatialreference and register it on the fly  `https://spatialreference.org/ref/{auth}/{code}/ogcwkt/`:     https://gist.github.com/mdsumner/fb8373927fe026cd8bf6bf6700179040

Hacky workaround for 9354 specifically (this would be a never-ending task): 

https://gist.github.com/mdsumner/a2115df9dd7c3aac3c353f1e784a725c#projection-string-bug

kylebarron:
CRS registration isn't required by this library. We do and should continue to expose generic functions for converting GeoTIFF geokeys to a proj4js converter. But it's up to the user whether that happens by fetching spatialreference.org or what.

PR #208, opened 2026-01-28
feat: Overhaul to use our `geotiff` package & generic TileMatrixSet support

### LINKED ISSUE #216: Use exact TileMatrixSet specification
In our current implementation here, we have a `TileMatrixSet` type that is very similar to the [OGC TileMatrixSet spec](https://github.com/opengeospatial/2D-Tile-Matrix-Set), but not exactly:
https://github.com/developmentseed/deck.gl-raster/blob/fa9e174720259f150a52a09b43a94b335046df33/packages/deck.gl-raster/src/raster-tileset/types.ts#L177-L204

It would be good to switch our implementation to use the exact `TileMatrixSet` type for interoperability.

I started creating a new port of [morecantile](https://github.com/developmentseed/morecantile) to TypeScript: [morecantile-ts](https://github.com/developmentseed/morecantile-ts). This has auto-generated types from the underlying spec. I'd like to put some helpers related to morecantile (like affine transforms) in morecantile-ts that we can then depend on here.

kylebarron:
This is a really big PR, but unfortunately it really had to be a big PR because we effectively had to change our geotiff reader _and_ our TileMatrixSet traversal at the same time.

### Change list

- Fully removes `geotiff.js` as a dependency. We now only use `@developmentseed/geotiff` and the underlying `@cogeotiff/core` for reading data.
- The `TileMatrixSetTileset` now accepts an _arbitrary_, _spec-compliant_, OGC TileMatrixSet instance. This will make it much more interoperable. Closes https://github.com/developmentseed/deck.gl-raster/issues/216. This should make it easier for Lonboard to generate a TMS from async-geotiff, pass the TMS definition to JS, and then have it auto-fetch tiles from Python according to that tiling system.

- Removes `GeoTIFFLayer` from our public API for now, since our geotiff implementation doesn't support fetching the entire full-resolution image and stitching tiles together. (And it's probably a candidate to fully remove in the future in favor of `COGLayer`)
- Define new projection resolver `epsgResolver`. Our default implementation caches all fetched projections in our `PROJECTION_REGISTRY` (closes https://github.com/developmentseed/deck.gl-raster/issues/185). But users can easily override this behavior by passing a custom `epsgResolver` 
- Removes multiple copies of affine handling, ensuring we put all affine transformation handling into `@developmentseed/affine`
- Update `cog-basic` example for latest changes. (Other examples haven't been updated yet)

`@cogeotiff/core` is _very_ pleasant to use. It's nice to use Typescript Enums instead of hard-coding integer matching.

### Todo

- [ ] Ensure request cancellation (needs https://github.com/blacha/cogeotiff/pull/1400 upstream). Should make the deck.gl visualization quite a bit faster.
- [ ] Move image decoding onto a thread pool in workers.

Closes #216

### DIFF
--- a/packages/deck.gl-geotiff/tsconfig.build.json
+++ b/packages/deck.gl-geotiff/tsconfig.build.json
@@ -8,8 +8,9 @@
   "include": ["src/**/*"],
   "exclude": ["node_modules", "dist", "tests"],
   "references": [
+    { "path": "../deck.gl-raster/tsconfig.build.json" },
     { "path": "../geotiff/tsconfig.build.json" },
-    { "path": "../raster-reproject/tsconfig.build.json" },
-    { "path": "../deck.gl-raster/tsconfig.build.json" }
+    { "path": "../morecantile/tsconfig.build.json" },
+    { "path": "../raster-reproject/tsconfig.build.json" }
   ]
 }
--- a/packages/deck.gl-raster/tsconfig.build.json
+++ b/packages/deck.gl-raster/tsconfig.build.json
@@ -7,5 +7,8 @@
   },
   "include": ["src/**/*"],
   "exclude": ["node_modules", "dist", "tests"],
-  "references": [{ "path": "../raster-reproject/tsconfig.build.json" }]
+  "references": [
+    { "path": "../morecantile/tsconfig.build.json" },
+    { "path": "../raster-reproject/tsconfig.build.json" }
+  ]
 }
--- a/packages/geotiff/tsconfig.build.json
+++ b/packages/geotiff/tsconfig.build.json
@@ -7,5 +7,8 @@
   },
   "include": ["src/**/*"],
   "exclude": ["node_modules", "dist", "tests"],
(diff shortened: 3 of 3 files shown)


packages/deck.gl-geotiff/src/cog-layer.ts
```diff
@@ -268,30 +267,27 @@ export class COGLayer<
...
     // Select overview image
-    const geotiffImage = images[images.length - 1 - z]!;
-    const imageHeight = geotiffImage.getHeight();
-    const imageWidth = geotiffImage.getWidth();
+    // If z=0, use the coarsest overview (which is the last in the array)
+    // If z=max, use the full-resolution image (which is the first in the array)
 
-    const tileMatrix = metadata.tileMatrices[z]!;
-    const { tileWidth, tileHeight } = tileMatrix;
+    // TODO: should be able to optimize this to not create the array
```
kylebarron:
```suggestion
    // TODO: should be able to (micro) optimize this to not create the array
```

PR #209, opened 2026-01-29
feat: zarr

Shane98c:
WIP.


kylebarron (commented):
First of all, thank you, this is awesome and I'm excited to dig in to it with you.

Next, this is just too big for a single PR. We should work to split this into more bite size PRs:

- Ideally we should be able to split any changes to deck.gl-geotiff outside of this PR

## Location for zarr-multiscale-metadata

I'm not sure if you have any opinions over where zarr-multiscale-metadata should live as a package? It seems general enough to live in a repo outside of a purpose-focused deck.gl rendering solution. 

I created https://github.com/developmentseed/deck.gl-raster/issues/210 in case you want to discuss this more outside of the context of this PR. Perhaps it would make sense to have a standalone `zarrita-multiscales` repo? Is there any appetite for moving it into zarrita directly?

packages/deck.gl-geotiff/src/cog-layer.ts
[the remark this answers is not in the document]
kylebarron: These changes could make sense, but best to discuss in a standalone PR I think

packages/deck.gl-geotiff/src/proj.ts
```diff
@@ -2,6 +2,12 @@ import type { ProjectionDefinition } from "proj4";
 import proj4 from "proj4";
 import type { PROJJSONDefinition } from "proj4/dist/lib/core";
 
+// Register EPSG:3857 (Web Mercator) - not included in proj4 by default
+proj4.defs(
+  "EPSG:3857",
+  "+proj=merc +a=6378137 +b=6378137 +lat_ts=0 +lon_0=0 +x_0=0 +y_0=0 +k=1 +units=m +nadgrids=@null +wktext +no_defs +type=crs",
+);
+
```
kylebarron: That's incorrect

Image

From http://proj4js.org/

packages/deck.gl-raster/src/gpu-modules/reproject-4326.ts
kylebarron: I need to think through this. 

1. Feels like some of this reprojection should be handled in the vertex shader rather than the fragment shader
2. The intent of the raster-reproject module was that we'd never have to do complex fragment shader reprojection like this.

I'd like to first support rendering without this GPU-based reprojection, and later on we can come back to this as an enhancement.

This means that we can go down two paths concurrently:

- Work on rendering COGs from EPSG:4326, to ensure that we're rendering EPSG:4326 data correctly. 
- Work on rendering Zarr from Mercator, Albers, or some other projection that is known to work on COG input data already.

Otherwise, there are too many moving parts and it's too complicated to do both of these at the same time.

packages/deck.gl-raster/src/raster-layer.ts
```diff
@@ -321,3 +526,85 @@ function reprojectorToMesh(reprojector: RasterReprojector): {
...
+ * WebMercatorViewport), no reprojection is needed. GPU rasterization interpolates
+ * UVs in screen space, which is Mercator space after projection, so sampling is
+ * linear in Mercator Y - matching the Mercator texture layout.
+ *
+ * Note: This assumes WebMercatorViewport. Non-Mercator viewports (GlobeView,
+ * OrthographicView) would require different handling.
+ *
+ * @param bounds Geographic bounds in WGS84
+ * @param latIsAscending Whether texture row 0 is south (affects UV mapping)
+ */
```
kylebarron: I think we could make some utilities in the `raster-reproject` package for generating a quad mesh. That could then be self contained and outside of the `RasterLayer`

packages/deck.gl-zarr/src/ellipsoids.ts
kylebarron: I'd like to not have this vendored _twice_ in the repo. Maybe it makes sense to have some sort of projection helpers defined somewhere that both `deck.gl-geotiff` and `deck.gl-zarr` can use. But I'd like for `deck.gl-raster` to not take a dependency on `proj4`

packages/deck.gl-zarr/tests/zarr-tile-matrix-set.test.ts
kylebarron: This would be a great place to start with bite-size Zarr PRs.

- We discuss and figure out the tileset abstraction we want to use for Zarr rendering. Is it trying to shoehorn into TileMatrixSet, or is it trying to use the bleeding edge GeoZarr/multiscales state of the art?
- Then we add parsers and test cases specifically for that tileset metadata. Then we can discuss and merge PRs without any sort of visualization yet; just with Zarr tileset traversal.
- At this stage, it would be great to have a Zarr "debug application", that just renders boxes for each Zarr chunk, printing the chunk id (like `0.3.2.4`, or something), so we can explore and visually verify how the Zarr traversal is working, before we touch any of the Zarr data rendering itself.

	Another nice thing here is that this isolates the Zarr chunk traversal with the Zarr chunk reprojection. We don't need to touch any of the EPSG:3857/EPSG:4326 rendering behavior. All we care about at this stage is verifying that chunks are chosen.
- Then we move onto Zarr reprojection (e.g. rendering a debug mesh)
- Then we move onto Zarr data rendering

kylebarron:
I've been reading back through this PR and I think the vast majority of this PR is now implemented on main. 

- There could be some elements of zarr multiscale metadata parsing that are still useful. I'm not sure exactly how to extract that. For now, I think the ZarrLayer on main will expect GeoZarr metadata, and it's up to the user (or helper functions in `@developmentseed/geozarr`) to infer GeoZarr metadata from the existing metadata.
- I don't _think_ any of the reprojection code is still relevant but I'm not 100% sure

I'm going to close this PR, but I welcome new PRs on latest main

Issue #217, opened 2026-02-05
Make higher-level GeoTIFF/COG abstraction similar to Python's async-geotiff


kylebarron:
We currently use geotiff.js as our underlying TIFF reader in the browser. Despite the name, it's really not a _Geo_-specific TIFF reader at all. It's hard to find a single [API it exports](https://geotiffjs.github.io/geotiff.js) that is specific to geospatial (other than parsing GeoTIFF metadata). 

This means it's much closer to our [_`async-tiff`_](https://github.com/developmentseed/async-tiff): geotiff.js is a low-level TIFF reader. 

But it's quite useful to have a higher-level abstraction, as we found while creating the Python [`async-geotiff`](https://github.com/developmentseed/async-geotiff). For example, geotiff.js doesn't do any special handling of mask IFDs, making it very easy to incorrectly render masks as normal IFDs (#196).

Following the API of async-geotiff, we should create higher-level GeoTIFF API, at least for internal use

PR #219, opened 2026-02-05
add hash params url= and iframe= to cog-basic example.

bdon:
* enables loading arbitrary COGs and a cleaner UI for embedding in an iframe.

Could eventually be embedded in source.coop: see https://github.com/source-cooperative/source.coop/pull/169


kylebarron:
Awesome! In the future we can revisit this, but I want to focus on hardening the library first.

I took a pause here for a week to build async-geotiff, which has given me a much better low level understanding of various different profiles of GeoTIFFs. Now I think especially with https://github.com/developmentseed/deck.gl-raster/issues/217 it'll help us make progress towards reliably rendering more types of COGs

PR #263, opened 2026-02-25
feat: add zstd via fzstd

gadomski:


packages/geotiff/src/fetch.ts
```diff
@@ -41,18 +41,12 @@ export async function fetchTile(
     throw new Error("Mask fetching not implemented yet");
   }
 
-  const tile = await self.image.getTile(x, y, options);
-  if (tile === null) {
-    throw new Error("Tile not found");
-  }
```
kylebarron: We might still want to have a relevant check for whether the tile/band being fetched is in the range of the image.

packages/geotiff/src/fetch.ts
```diff
@@ -96,6 +113,71 @@ export async function fetchTile(
...
+async function fetchBandSeparateTile(
+  self: HasTiffReference,
+  x: number,
+  y: number,
+  opts: {
+    sampleFormat: SampleFormat;
+    bitsPerSample: number;
+    samplesPerPixel: number;
+    predictor: CachedTags["predictor"];
+    planarConfiguration: CachedTags["planarConfiguration"];
```
kylebarron: I think here it would be cleaner if you just used the `Predictor` and `PlanarConfiguration` types directly. It's not as obvious with this indirection what `CachedTags["predictor"]` is defined as

gadomski:
> can you split this PR into two? One for adding ZSTD and a separate PR for loading band-separate data?

Done: https://github.com/developmentseed/deck.gl-raster/pull/276

packages/geotiff/README.md
```diff
@@ -75,6 +75,19 @@ Until you try to load an image compressed with, say, [LERC], you don't pay for t
 
 [LERC]: https://github.com/Esri/lerc
 
+You can also override the built-in decoders with your own by using `registry`. For example, to use a custom zstd decoder:
```
kylebarron: 🎉

PR #266, opened 2026-02-25
feat: Globe view, antimeridian, and polar projection support

espg:
This is in response to the discussion in [this issue](https://github.com/source-cooperative/source.coop/issues/230) , about the possibility of defaulting to a global viewer in the source.coop stac-map viewer.

## Summary

- Implement globe view tile traversal by threading `projectTo4326` through the tileset and computing bounding volumes in 3D globe common space via `viewport.projectPosition`
- Add antimeridian crossing detection in `RasterReprojector` — when a tile's WGS84 longitudes span >180°, normalize them to a continuous range (e.g. [170, 190] instead of [170, -170]) so mesh triangles don't wrap the wrong way around the globe
- Add polar projection support with pole detection and a `maxTriangles` safety cap on `run()` to prevent infinite refinement near poles
- Add 37 new tests across tile traversal, antimeridian handling, and polar projections (up from 1 placeholder test)
- Add a `globe-view` example using `GlobeView` + `COGLayer`

Closes #82, closes #171, closes #172

## Commits

1. **Baseline tests** — 21 tests for tile traversal code (bounds, LOD, child ranges, OBB) establishing a regression baseline before changes
2. **Globe view traversal** — `_getGlobeBoundingVolume()`, `sampleReferencePointsInWgs84()`, globe-mode bounds conversion in `getTileIndices()`, LOD fix using `centerLatitude` instead of Mercator-specific `worldToLngLat()`
3. **Antimeridian** — detect ±180° crossing from initial corner vertices, shift negative longitudes by +360, apply offset to subsequent vertices during mesh refinement
4. **Polar projections** — detect tiles near poles (|lat| > 75° + antimeridian crossing), `maxTriangles` parameter on `run()` to cap refinement
5. **Globe example** — minimal `GlobeView` + `COGLayer` example with debug controls

## Test plan

- [x] All 42 unit tests pass (`pnpm test` in `deck.gl-raster` and `raster-reproject`)
- [x] Run `globe-view` example with NZ imagery URL — verify tiles render on the globe
- [x] Run `cog-basic` example — verify no regression in Mercator rendering

PR #268, opened 2026-02-26
feat: Add globe view support for tile traversal

### LINKED ISSUE #82: Support for rendering in globe view
I'm not sure whether a `RasterLayer` would work out of the box, but I know a `COGLayer` definitely would not. We should test whether a `RasterLayer` works in GlobeView and we should try to [follow upstream code](https://github.com/visgl/deck.gl/blob/a15c8cea047993c8a861bf542835c1988f30165c/modules/geo-layers/src/tileset-2d/tile-2d-traversal.ts#L149-L170) to try to get our generic tiling approach to work in the GlobeView.

espg:
Ported from #266.

## Summary

- Implement globe view tile traversal by threading `projectTo4326` through the tileset and computing bounding volumes in 3D globe common space via `viewport.projectPosition`
- Add `_getGlobeBoundingVolume()` and `sampleReferencePointsInWgs84()` for globe-mode bounds conversion in `getTileIndices()`
- Fix LOD calculation to use `centerLatitude` instead of Mercator-specific `worldToLngLat()`
- Add 25 new unit tests for tile traversal (bounds, LOD, child ranges, OBB) — up from 1 placeholder test
- Add a `globe-view` example demonstrating `GlobeView` + `COGLayer` with debug controls

This is in response to the discussion in [source-cooperative/source.coop#230](https://github.com/source-cooperative/source.coop/issues/230), about the possibility of defaulting to a global viewer in the source.coop stac-map viewer.

Closes #82

## Test plan

- [x] All unit tests pass (`pnpm test` in `deck.gl-raster`)
- [x] Run `globe-view` example with NZ imagery URL — verify tiles render on the globe
- [x] Run `cog-basic` example — verify no regression in Mercator rendering

### DIFF
--- a/examples/globe-view/index.html
+++ b/examples/globe-view/index.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>COGLayer Globe View Example</title>
+    <style>
+      body {
+        margin: 0;
+        font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', sans-serif;
+      }
+      #root {
+        width: 100vw;
+        height: 100vh;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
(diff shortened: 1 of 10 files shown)


espg:
It looks like these are the standard projection issues that happen when trying to draw straight lines that interpolate from lat / lon corners without drawing them as curved arcs. The quick fix is to [densify edges](https://github.com/developmentseed/deck.gl-raster/blob/main/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L458) prior to projection. I went ahead and added this, but don't have a good test case handy to verify it with

kylebarron:
This is now superseded by #563

PR #305, opened 2026-03-04
feat: Return ImageBitmap in canvas decoding

### LINKED ISSUE #228: (geotiff): Allow returning JPEG or WebP-compressed GeoTIFF tiles as ImageBitmap
In the future, we could have an API that returns an `ImageBitmap` directly from the decoder, to avoid copying pixel data from GPU -> CPU memory.

Then deck.gl could use the ImageBitmap directly as a texture source without copying again from CPU -> GPU memory.

kylebarron:
Closes https://github.com/developmentseed/deck.gl-raster/issues/228. I'm not sure the juice is worth the squeeze here, at least not until we're really focused on performance optimization.


kylebarron:
This is now stale, and all of this is just to avoid like one image data copy, so for now I'm going to close this. Ref #621

PR #321, opened 2026-03-06
fix: Handle visualizing COGs with flipped y (origin at the bottom left)

### LINKED ISSUE #303: Support flipped-y images (origin at the bottom left, not top left)
To unblock #261 because the raw AEF GeoTIFFs were exported upside down.

@gadomski put a test image in geotiff-test-data https://github.com/developmentseed/geotiff-test-data/blob/main/real_data/source-coop-alpha-earth/xjejfvrbm1fbu1ecw-0000000000-0000008192_info.md

kylebarron:
We should still make sure that the _data_ is not flipped when visualizing

Closes https://github.com/developmentseed/deck.gl-raster/issues/303


kylebarron:
This is stale, and I don't think we have an immediate need to render flipped COGs, so I'm going to close this out in service of #621. 

But we can note on #303 that a future PR can evaluate the changes prototyped on this branch

PR #348, opened 2026-03-18
docs: add-plausible

aboydnw:
I created a new site in plausible and added the custom script here.


kylebarron (approved):

kylebarron:
@aboydnw just realized that this doesn't have metrics on the published examples, which are deployed as independent static sites. Not sure if you want to change that?

PR #354, opened 2026-03-18
fix: Support TileLayer refinement strategies

### LINKED ISSUE #190: Implement custom Refinement Strategy in `COGLayer`
When zooming in, we should render based on the lower-resolution data until the higher resolution data loads.

See https://deck.gl/docs/api-reference/geo-layers/tile-layer#refinementstrategy

Even though the default value of that is `best-available`, I figure that their implementation is probably tied to Web Mercator indexing, and since we don't have web mercator indexing, we probably have to implement a custom `refinementStrategy`

cc @batpad

kylebarron:
### Change list

- Define `TileMetadata` type in `raster-tileset-2d` for object returned by `Tileset.getTileMetadata`
- Return `bbox` object from `getTileMetadata`, which is the WGS84 bounding box (edge densified) of the current tile.
- Fix decimation calculation to consider both image size _and_ cell size. For example, sentinel 2 has a different image size for its overviews (512x512) as for the full-resolution data (1024x1024). We need to handle this offset correctly.
- Add tests 

https://github.com/user-attachments/assets/1e71273d-4cd3-4b6e-9073-7ae00267a2b9

Closes #190, closes #192

### DIFF
--- a/packages/deck.gl-raster/tests/tileset-refinement.test.ts
+++ b/packages/deck.gl-raster/tests/tileset-refinement.test.ts
@@ -7,6 +7,7 @@
 
 import type { Viewport } from "@deck.gl/core";
 import type { _Tileset2DProps as Tileset2DProps } from "@deck.gl/geo-layers";
+import type { TileMatrixSet } from "@developmentseed/morecantile";
 import { describe, expect, it } from "vitest";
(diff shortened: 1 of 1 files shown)


edeetee:
Thank you legend!

PR #374, opened 2026-03-25
fix: handle antimeridian-crossing tiles in Web Mercator reprojection

### LINKED ISSUE #366: COG tiles near ±180° longitude cause RasterReprojector mesh divergence in Web Mercator mode
## Description

When rendering an EPSG:4326 COG that spans the antimeridian (±180° longitude) in a Web Mercator viewport, tiles near the dateline cause `RasterReprojector` mesh refinement to diverge, producing mislocated rectangles across the map.

The root cause: `forwardTo3857` maps nearby source coordinates on opposite sides of ±180° to EPSG:3857 x-values ~40 million meters apart (+20M vs -20M). The mesh triangles connecting these vertices span the entire map, and the reprojection error never converges (error=43200 pixels after 10001 iterations).

## Reproduction

Load a global EPSG:4326 COG (e.g. WorldPop population data):

```ts
new COGLayer({
  geotiff: 's3://wherobots-examples/data/ppp_2020_1km_Aggregated.tif',
  // ...
})
```

Console shows:

```text
RasterReprojector: mesh refinement did not converge after 10001 iterations
(maxError=0.125, currentError=43200.00017279999)
```

## Analysis

In `_renderSubLayers`, for Web Mercator viewports, `reprojectionFns.forwardReproject` is set to `forwardTo3857`. For tiles at the antimeridian, this projection has a discontinuity — longitude values like +179° and -179° (which are 2° apart geographically) map to x-values ~40M meters apart in EPSG:3857. The RasterReprojector mesh subdivision cannot resolve this discontinuity no matter how many iterations it runs.

## Workaround

We're patching `_renderSubLayers` to detect antimeridian-crossing tiles by checking if the tile's corner x-coordinates in EPSG:3857 span more than half the world circumference. When detected, `forwardTo3857` is wrapped to shift negative x-values by +circumference, making the mesh continuous:

```ts
const HALF_CIRCUMFERENCE = WEB_MERCATOR_METER_CIRCUMFERENCE / 2;
const cornerXs = /* compute tile corners in EPSG:3857 */;
if (Math.max(...cornerXs) - Math.min(...cornerXs) > HALF_CIRCUMFERENCE) {
    tileForwardTo3857 = (x, y) => {
        const [px, py] = forwardTo3857(x, y);
        return [px < 0 ? px + WEB_MERCATOR_METER_CIRCUMFERENCE : px, py];
    };
    tileInverseFrom3857 = (x, y) => {
        const unwrapped = x > HALF_CIRCUMFERENCE ? x - WEB_MERCATOR_METER_CIRCUMFERENCE : x;
        return inverseFrom3857(unwrapped, y);
    };
}
```

## Environment

`@developmentseed/deck.gl-geotiff`: `0.4.0`
`@developmentseed/raster-reproject`: `0.4.0`

james-willis:
## Summary

Fixes #366 — COG tiles near ±180° longitude cause `RasterReprojector` mesh refinement to diverge in Web Mercator mode.

## Problem

When rendering an EPSG:4326 COG that spans the antimeridian, `forwardTo3857` maps nearby source coordinates on opposite sides of ±180° to EPSG:3857 x-values ~40 million meters apart (+20M vs -20M). The mesh triangles connecting these vertices span the entire map, and the reprojection error never converges (`currentError=43200` after 10001 iterations).

## Solution

In `COGLayer._renderSubLayers`, for the Web Mercator code path:

1. **Detect antimeridian-crossing tiles** — project the four tile corners to EPSG:3857 and check whether their x-values span more than half the world circumference (~20M meters)
2. **Wrap the projection functions** — shift negative EPSG:3857 x-values by `+circumference` so the tile's coordinate space is continuous (all positive), and unwrap in the inverse direction

This makes the mesh subdivision see a smooth, continuous coordinate space instead of a discontinuity, allowing `RasterReprojector` to converge normally.

## Changes

- `packages/deck.gl-geotiff/src/cog-layer.ts`:
  - Added `HALF_CIRCUMFERENCE` constant
  - Added antimeridian detection + projection wrapping logic in the Web Mercator branch of `_renderSubLayers`

## Testing

- Build passes (`pnpm build`)
- Lint passes (`pnpm lint`)
- Existing tests are unaffected (pre-existing Node.js version issue prevents test runner from starting on this machine)

### DIFF
--- a/packages/deck.gl-geotiff/src/cog-layer.ts
+++ b/packages/deck.gl-geotiff/src/cog-layer.ts
@@ -39,7 +39,11 @@ import type { TextureDataT } from "./geotiff/render-pipeline.js";
 import { inferRenderPipeline } from "./geotiff/render-pipeline.js";
 import { fromAffine } from "./geotiff-reprojection.js";
 import type { EpsgResolver } from "./proj.js";
-import { epsgResolver, makeClampedForwardTo3857 } from "./proj.js";
+import {
+  epsgResolver,
+  makeClampedForwardTo3857,
+  wrapAntimeridianProjections,
+} from "./proj.js";
 
 /** Size of deck.gl's common coordinate space in world units.
  *
@@ -463,11 +467,29 @@ export class COGLayer<
         };
         deckProjectionProps = {};
       } else {
+        // Wrap projection fns for antimeridian-crossing tiles so the 3857
+        // x-coordinates are continuous and RasterReprojector can converge (#366).
+        const corners = [
+          [0, 0],
+          [width, 0],
+          [width, height],
+          [0, height],
+        ] as const;
+        const cornerXs = corners.map(([cx, cy]) => {
+          const [sx, sy] = forwardTransform(cx, cy);
+          return forwardTo3857(sx, sy)[0];
+        });
+        const wrapped = wrapAntimeridianProjections(
+          cornerXs,
+          forwardTo3857,
+          inverseFrom3857,
+        );
+
         reprojectionFns = {
           forwardTransform,
           inverseTransform,
-          forwardReproject: forwardTo3857,
-          inverseReproject: inverseFrom3857,
+          forwardReproject: wrapped.forwardTo3857,
+          inverseReproject: wrapped.inverseFrom3857,
         };
         // Scale 3857 meters → deck.gl world units (512×512).
         //
--- a/packages/deck.gl-geotiff/src/proj.ts
+++ b/packages/deck.gl-geotiff/src/proj.ts
@@ -8,6 +8,11 @@ const WGS84_ELLIPSOID_A = 6378137;
 // Beyond this, the Mercator projection is undefined.
 const MAX_WEB_MERCATOR_LAT = 85.05112877980659;
 
+const WEB_MERCATOR_METER_CIRCUMFERENCE = 2 * Math.PI * WGS84_ELLIPSOID_A;
+const HALF_CIRCUMFERENCE = WEB_MERCATOR_METER_CIRCUMFERENCE / 2;
+
+type ProjectionFn = (x: number, y: number) => [number, number];
+
 /**
  * Convert a WGS84 longitude/latitude to EPSG:3857 meters analytically.
  * Valid for latitudes in [-MAX_WEB_MERCATOR_LAT, MAX_WEB_MERCATOR_LAT].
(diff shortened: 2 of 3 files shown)


packages/deck.gl-geotiff/src/proj.ts
```diff
@@ -8,6 +8,11 @@ const WGS84_ELLIPSOID_A = 6378137;
 // Beyond this, the Mercator projection is undefined.
 const MAX_WEB_MERCATOR_LAT = 85.05112877980659;
 
+const WEB_MERCATOR_METER_CIRCUMFERENCE = 2 * Math.PI * WGS84_ELLIPSOID_A;
+const HALF_CIRCUMFERENCE = WEB_MERCATOR_METER_CIRCUMFERENCE / 2;
+
+type ProjectionFn = (x: number, y: number) => [number, number];
```
kylebarron: Edit you can use the copy in `transform-bounds.ts`

packages/deck.gl-geotiff/src/cog-layer.ts
```diff
@@ -463,11 +467,29 @@ export class COGLayer<
...
+        ] as const;
+        const cornerXs = corners.map(([cx, cy]) => {
+          const [sx, sy] = forwardTransform(cx, cy);
+          return forwardTo3857(sx, sy)[0];
+        });
+        const wrapped = wrapAntimeridianProjections(
+          cornerXs,
+          forwardTo3857,
+          inverseFrom3857,
+        );
```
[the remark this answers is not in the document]
kylebarron: This does seem like a valid critique. I'm not sure at a glance how it could be fixed.

packages/deck.gl-geotiff/src/proj.ts
```diff
@@ -19,6 +24,37 @@ function wgs84To3857(lon: number, lat: number): [number, number] {
   return [x, y];
 }
 
+/**
+ * If a tile's EPSG:3857 corner x-values span more than half the globe, wrap
+ * `forwardTo3857` / `inverseFrom3857` so the coordinate space is continuous.
+ *
+ * Returns the original functions unchanged when no wrapping is needed.
+ */
```
gadomski, line 32: Would this heuristic trigger on global rasters?

Since we have the corners of the raster in WGS84, can't we instead just check to see whether the left longitude is greater than the right longitude?

kylebarron: Thanks @gadomski, this is a good point. 

If we have an image that's 512px wide, 256px tall that covers the entire EPSG:4326 coordinate space, then it would trigger this function, even though it doesn't wrap the antimeridian.

(We should add unit test cases that such an input image doesn't trigger antimeridian handling)

PR #444, opened 2026-04-20
feat: temperature anomaly deck.gl-raster zarr example

wrynearson:
## What I am changing

- Add example with temperature anomaly zarr
- Added dropdown to select data variable
- Query data values on click

## How I did it

- Created a new branch from kyle/zarr-dynamical-ecmwf
- Relied on help from Claude

## How you can test it

- In this branch, in /examples/temp-anomaly/src, run `pnpm dev`

## Related Issues

https://github.com/user-attachments/assets/7f59f562-9345-438e-9e97-3d485bb8295a

### DIFF
--- a/dev-docs/specs/2026-04-17-ecmwf-zarr-animation-design.md
+++ b/dev-docs/specs/2026-04-17-ecmwf-zarr-animation-design.md
@@ -1,0 +1,256 @@
+# ECMWF Zarr Animation Example — Design
+
+## Problem
+
+The new `examples/dynamical-zarr-ecmwf` example needs to visualize a single
+scalar variable from the ECMWF IFS ENS 15-day forecast dataset hosted by
+Dynamical.org. The dataset is a Zarr v3 store that is **not** GeoZarr-compliant,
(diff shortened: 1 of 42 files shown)


kylebarron:
@wrynearson do I understand correctly that you're publishing this as a separate website? Should we close this PR?

kylebarron:
@wrynearson I think it would be great to have a "showcase" page in the docs, thinking of 

https://deck.gl/showcase

or https://developmentseed.org/lonboard/latest/examples/#third-party-showcase

if you'd want to publicize this or any other project, we could add it. I think I'd lean towards publicizing a _real app_ that you've built and deployed to production, and I'm not sure this example is that

wrynearson:
Thanks for the offer @kylebarron. Agreed that I don't think this would qualify: I mostly re-used your code when I branched off of one of your examples to make this 😄.

I'll continue to work on https://developmentseed.org/weather-anomaly/ separately. If it gets to the state of a _real all_, we can re-assess adding it as a showcase.

PR #445, opened 2026-04-20
feat: Zarr temporal animation

### LINKED ISSUE #403: Zarr: AbortSignal support
AbortSignal support added in upcoming release in https://github.com/manzt/zarrita.js/pull/379

kylebarron:
https://github.com/user-attachments/assets/8d8979ad-8abf-4776-9502-deeda43bfd1e

### Change list

- Update `ZarrLayer`:
	- Force users to open the zarrita store themselves. They must pass in a `zarr.Array` or `zarr.Group`
	- Force users to pass a [`selection`](https://zarrita.dev/packages/zarrita.html#slicing-and-indexing) for all non-spatial dimensions.
		- Validate that user has explicitly selected all non-spatial dimensions. 
		- For now, the contract is that deck.gl will automatically select spatial dimensions, and will automatically select across multiscale arrays (groups?). But the user must explicitly select all non-spatial dimensions. If they want an entire dimension, they must still explicitly pass `dimensionName: null`.
- Fix raster-tileset to clamp wgs84bounds to web mercator range.
- Create example app to animate ECMWF Zarr data
	- UI offers a choice of colormap options from the default Matplotlib colormap sprite added in #447 
	- The `getTileData` loads chunks into a `Texture2dArray` GPU texture, which has height, width, plus the depth of the number of time chunks in the given forecast.
	- Hard-code GeoZarr attributes
	- Render pipeline which applies filtering, linear rescaling, colormap
	- Selection that fetches the Zarr chunk at the given `init_time` and `ensemble_member` (defaulting to `0`)
	- GPU modules to filter within a range and load a texture2darray (these will go into library code in the future, but for now live as part of the example)
	- claude-written UI control panel, enabling animation

Closes https://github.com/developmentseed/deck.gl-raster/issues/403


packages/deck.gl-raster/src/raster-tileset/raster-tileset-2d.ts
```diff
@@ -82,10 +82,22 @@ export class RasterTileset2D extends Tileset2D {
...
+    // asserts against that range. Global data at ±90° (e.g. reanalysis grids)
+    // would otherwise crash tile selection. Clamp here; any polar rows beyond
+    // ±MAX_LAT are unreachable on a Mercator map anyway.
+    const MAX_LAT = 85.0511287798066;
+    this.wgs84Bounds = [
+      rawBounds[0],
+      Math.max(rawBounds[1], -MAX_LAT),
+      rawBounds[2],
+      Math.min(rawBounds[3], MAX_LAT),
+    ];
```
kylebarron, line 100: I think this can be split out into a separate PR fix

examples/dynamical-zarr-ecmwf/src/App.tsx
```diff
@@ -0,0 +1,234 @@
...
+  );
+
+  // Decode the shipped colormap sprite once at mount. This returns ImageData
+  // and doesn't need a GPU device, so it can run in parallel with zarr
+  // opening.
+  const [colormapImage, setColormapImage] = useState<ImageData | null>(null);
+  useEffect(() => {
+    let cancelled = false;
+    (async () => {
+      const bytes = await (await fetch(colormapsPngUrl)).arrayBuffer();
```
kylebarron: nit: move into two lines for readability (I hate `await await`)
```
      const bytes = await (await fetch(colormapsPngUrl)).arrayBuffer();
```

examples/dynamical-zarr-ecmwf/src/App.tsx
```diff
@@ -0,0 +1,234 @@
...
+      cancelled = true;
+    };
+  }, []);
+
+  // The colormap 2d-array texture is created lazily on the first tile load
+  // (when the luma.gl Device becomes available via options.device). We store
+  // it in a ref, not state, because the first renderTile call for the first
+  // tile happens synchronously after its getTileData resolves, and deck.gl
+  // guarantees getTileData → renderTile ordering for a given tile.
+  const colormapRef = useRef<Texture | null>(null);
```
kylebarron: Instead we can grab the device from `onDeck` initialized I think

kylebarron: Fixed in [`b9289a2` (this PR)](https://github.com/developmentseed/deck.gl-raster/pull/445/commits/b9289a28853be29d9843233185f9a2088569d2e6)

packages/geozarr/src/parse.ts
```diff
@@ -41,13 +41,19 @@ export function parseGeoZarrMetadata(attrs: unknown): GeoZarrMetadata {
...
+  const X_LABELS = new Set(["x", "longitude", "lon"]);
+  const yAxisIndex = axes.findIndex((a) => Y_LABELS.has(a.toLowerCase()));
+  const xAxisIndex = axes.findIndex((a) => X_LABELS.has(a.toLowerCase()));
 
   if (yAxisIndex === -1 || xAxisIndex === -1) {
     throw new Error(
-      `spatial:dimensions must contain "x" and "y" (case-insensitive), got: ${JSON.stringify(axes)}`,
+      `spatial:dimensions must contain a Y axis ("y"/"latitude"/"lat") and ` +
+        `an X axis ("x"/"longitude"/"lon") (case-insensitive), got: ` +
+        `${JSON.stringify(axes)}`,
```
kylebarron, line 56: Not sure this is correct, but to be followed up in https://github.com/developmentseed/deck.gl-raster/issues/451

PR #468, opened 2026-04-24
chore: Update to uuid v14

jcoyne:
Versions of uuid prior to 14.0.0 have a vulnerability: https://github.com/advisories/GHSA-w5hq-g745-h8pq

### DIFF
--- a/packages/geotiff/package.json
+++ b/packages/geotiff/package.json
@@ -61,6 +61,6 @@
     "@developmentseed/morecantile": "workspace:^",
     "fzstd": "^0.1.1",
     "lerc": "^4.1.1",
-    "uuid": "^13.0.0"
+    "uuid": "^14.0.0"
   }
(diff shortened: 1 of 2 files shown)


kylebarron:
Thanks; if you could run `pnpm install` to regenerate the lockfile that would fix CI

kylebarron:
Also fwiw the vulnerability is irrelevant for us because we use uuid v4

https://github.com/developmentseed/deck.gl-raster/blob/d3ab42a2f3e8d18d35443097136ed7f8056dabf0/packages/geotiff/src/tile-matrix-set.ts#L11

The vulnerability is not effective on v4 according to the advisory

jcoyne:
Yes. I was just hoping to remove the spurious npm audit callout about the vulnerability

kylebarron:
Thanks

PR #469, opened 2026-04-24
feat: Titiler .npy tile example with RasterTileLayer

### LINKED ISSUE #461: Create example reading from titiler backend
We should be able to use the new `RasterTileLayer` concept (https://github.com/developmentseed/deck.gl-raster/issues/404) where we read in initial metadata, align the client TMS with whatever TMS titiler uses to provide tiles, and then it should be a pretty simple example.

cc @vincentsarago

kylebarron:
image

Validated that this does render from titiler.xyz

Closes https://github.com/developmentseed/deck.gl-raster/issues/461

## Summary

- New example at `examples/titiler-cog/` that renders tiles fetched from `titiler.xyz` as `.npy` numpy arrays, decoded client-side and drawn via `RasterTileLayer`.
- First example in the repo that uses `RasterTileLayer` directly (not via `COGLayer` / `ZarrLayer`). Demonstrates the flow for any server that hands back per-tile binary arrays for an OGC tile matrix set.
- Startup fetches `/cog/WebMercatorQuad/tilejson.json` (WGS84 bounds + min/max zoom) and `/tileMatrixSets/WebMercatorQuad` (tile pyramid). Builds a `TileMatrixSetAdaptor` with identity EPSG:3857 projections and proj4-backed EPSG:4326 projections, injects a `boundingBox` on the TMS, and fits the map to the dataset bounds.
- Per tile: `fetch .npy` → decode with `npyjs` → repack band-separate uint8 `(B, H, W)` to interleaved RGBA → upload RGB texture + optional `r8unorm` mask texture → pipeline `CreateTexture` + `MaskTexture`.

Design doc: [`dev-docs/specs/2026-04-24-titiler-cog-example-design.md`](dev-docs/specs/2026-04-24-titiler-cog-example-design.md).

## Test plan

- [ ] `pnpm --filter deck.gl-titiler-cog-example dev` renders the Sentinel-2 TCI RGB tiles over the dark basemap and fits the map to the scene.
- [ ] Zooming and panning loads new tiles from `titiler.xyz`.
- [ ] "Show Debug Mesh" toggle + "Debug Opacity" slider behave the same as in other examples.
- [ ] `pnpm --filter deck.gl-titiler-cog-example typecheck` passes.
- [ ] `pnpm --filter deck.gl-titiler-cog-example build` produces a working production bundle.


vincentsarago:
@kylebarron do you think https://github.com/developmentseed/titiler/issues/1242 could be of any help, like having the number of bands or the datatype ...

dev-docs/specs/2026-04-24-titiler-cog-example-design.md
```diff
@@ -0,0 +1,215 @@
...
+Uses the public `https://titiler.xyz` instance. Two endpoints:
+
+- `GET /cog/info?url=<COG_URL>` — returns metadata including
+  `bounds: [west, south, east, north]` in WGS84. Used to fit the map.
+- `GET /tileMatrixSets/WebMercatorQuad` — returns an OGC TMS 2.0 JSON
+  document. Used to build the `TilesetDescriptor`.
+- `GET /cog/tiles/WebMercatorQuad/{z}/{x}/{y}.npy?url=<COG_URL>` — returns
+  a numpy `.npy` file for one tile, shape `(bands, height, width)`,
+  dtype `uint8` for an RGB COG. For a 3-band RGB source titiler returns
+  4 bands: R, G, B, mask.
```
vincentsarago, line 38: where does this limitation comes from? Titiler could server tiles with nBands+masks

vincentsarago: Might be easier to support NPZ files which are binary files with one or two arrays (data and mask) https://github.com/cogeotiff/rio-tiler/blob/ebacf0aeee47f44c2f321bd0a362212a374813ee/rio_tiler/utils.py#L615-L622

https://numpy.org/doc/stable/reference/generated/numpy.savez.html

dev-docs/specs/2026-04-24-titiler-cog-example-design.md
```diff
@@ -0,0 +1,215 @@
...
+- Reactive URL / parameter changes at runtime. The COG URL is static.
+- Unit tests. Examples in this repo are not tested.
+
+## Server
+
+Uses the public `https://titiler.xyz` instance. Two endpoints:
+
+- `GET /cog/info?url=<COG_URL>` — returns metadata including
+  `bounds: [west, south, east, north]` in WGS84. Used to fit the map.
+- `GET /tileMatrixSets/WebMercatorQuad` — returns an OGC TMS 2.0 JSON
```
kylebarron, line 33: Also, we should keep in mind the difference between a:

- `RasterTileLayer`: our concept, which supports arbitrary tile grids
- upstream `TileLayer`: an official deck.gl layer which only supports the Web Mercator tiling grid but is a bit more optimized.

If a user only wants to render in web mercator, they can use the upstream `TileLayer` directly (with our `RasterLayer` as the generated sub layer).

@vincentsarago are there specific times when rendering in non-web mercator would be faster? Like if we rendered from the specific UTM zone a Sentinel tile was generated, for example?

dev-docs/specs/2026-04-24-titiler-cog-example-design.md
kylebarron: This limitation isn't "required"; it was just a simplifying assumption to get some sort of titiler example working. 

NPZ would be fine; we'd just have to unzip the files

kylebarron:
> @kylebarron do you think [developmentseed/titiler#1242](https://github.com/developmentseed/titiler/issues/1242) could be of any help, like having the number of bands or the datatype ...

First, take this implementation with a grain of salt. We don't _have_ to use the `tilejson` endpoint in this app if there's a different endpoint that makes more sense.

I'm not sure I love overloading the tilejson with that other metadata, but perhaps it's better than making two different requests

dev-docs/specs/2026-04-24-titiler-cog-example-design.md
vincentsarago: yeah it will be faster to `reduce` the re-projection pipeline, so if you choose a close CRS to the dataset CRS it should be faster (in theory)

vincentsarago:
> > First, take this implementation with a grain of salt. We don't have to use the tilejson endpoint in this app if there's a different endpoint that makes more sense.
>
> I'm not sure I love overloading the tilejson with that other metadata, but perhaps it's better than making two different requests

Yeah I'm not super fan as well but at the moment there is no solution, both /tiles/{tileMatrixSetId} and /tilejson.json do not send information about the data they are giving access to. Application that does client side rendering will have to know in advance how many bands they will have in each tile and what is the datatype (at minimum). If you you the `/info` endpoint you get this information but they you also need to call the one of the previous endpoint to get the TMS coverage (bounds, min/max zoom)

kylebarron:
Also linking @hrodmn 's branch on https://github.com/developmentseed/titiler-cmr-browser/tree/feat/deck.gl-raster, which is titiler-based.

PR #470, opened 2026-04-24
chore: Enforce `.js` extensions on relative imports

kylebarron:
## Summary

- Enables Biome's `correctness/useImportExtensions` as an error.
- Configures `extensionMappings` to rewrite `.ts`/`.tsx` → `.js` and `.mts`/`.cts` → `.mjs`/`.cjs`. Does NOT rewrite `.json` imports.
- Excludes `docs/api-sidebars/` from Biome (generated files).
- Autofix appended `.js` (or `.mjs`/`.cjs`) to every relative import across packages and examples.

Targeting `titiler-example` rather than `main` so the titiler-cog example files get the extensions from the same change.

## Test plan

- [ ] `pnpm -r --filter '!docs' typecheck` passes (docs has a pre-existing `baseUrl` deprecation error unrelated to this change).
- [ ] `pnpm biome check .` passes with no remaining diagnostics.
- [ ] Spot-check a few files (e.g. `packages/morecantile/src/tile.ts`, `examples/cog-basic/src/App.tsx`) to confirm extensions look right.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -20,7 +20,8 @@
       "!!packages/geozarr/geo-proj",
       "!!packages/geozarr/multiscales",
       "!!packages/geozarr/spatial",
-      "!!packages/morecantile/spec"
+      "!!packages/morecantile/spec",
+      "!!docs/api-sidebars"
     ]
   },
   "formatter": {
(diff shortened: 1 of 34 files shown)


kylebarron:
Superseded by https://github.com/developmentseed/deck.gl-raster/pull/472

PR #471, opened 2026-04-24
perf: use source.coop proxy for dynamical Zarr example

alukach:
### Change list

- Switching to source.coop proxy allows us to use HTTP/3 connections, which improves load times because we can use HTTP multiplexing and increase the max number of concurrent requests
- Set `maxRequests` higher than the default of `6`

### DIFF
--- a/examples/dynamical-zarr-ecmwf/src/App.tsx
+++ b/examples/dynamical-zarr-ecmwf/src/App.tsx
@@ -38,7 +38,7 @@ import { ControlPanel } from "./ui/control-panel.js";
 // before setting INIT_TIME_IDX below.
 // Direct S3 link is faster than dynamical.org proxy
 const ZARR_URL =
-  "https://s3.us-west-2.amazonaws.com/us-west-2.opendata.source.coop/dynamical/ecmwf-ifs-ens-forecast-15-day-0-25-degree/v0.1.0.zarr";
+  "https://data.source.coop/dynamical/ecmwf-ifs-ens-forecast-15-day-0-25-degree/v0.1.0.zarr";
 
 const VARIABLE = "temperature_2m";
 const ENSEMBLE_MEMBER_IDX = 0; // control run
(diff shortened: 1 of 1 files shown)


examples/dynamical-zarr-ecmwf/src/App.tsx
```diff
@@ -225,6 +225,7 @@ export default function App() {
             selection,
             getTileData,
             renderTile,
+            maxRequests: 0,
```
kylebarron: This should still be a positive number, not 0, or else request cancellation is _never_ applied. 

I.e. if you pan around the world instantly, all of those requests that were visible for a fraction of a second will still be allowed to finish. 

I think `20` is a reasonable suggestion

kylebarron:
```suggestion
            maxRequests: 20,
```

alukach: It surprises me that this effects the behavior of cancelling requests that are out of the viewport, but if so then I agree

kylebarron: Hmm, perhaps I spoke too soon. I can't remember exactly how deck.gl's request cancellation works

kylebarron: I just verified that my initial hunch was right. 

**If you set `maxRequests: 0` you will never get request cancellation.** When set to `maxRequests: 0`, I tested panning very fast and I never saw a request get cancelled. With 20 seconds of fast panning I requested 3GB of data chunks 😄 

I think this is because the request scheduler works by checking:

- What are the tiles we have yet to request?
- When is there new slots to request tiles

I guess it's essentially a semaphore?

When `maxRequests: 0`, the request has already been sent, so deck.gl doesn't see it to cancel it...?

I'm not sure this makes sense, but it's what I see in practice.

PR #478, opened 2026-04-24
docs: inject Plausible analytics into example sites via CI

aboydnw:
This adds a CI step that injects the same Plausible script already used on the docs site into every example index.html after build. New examples added in the future are covered automatically with no additional changes required.

To verify, after deploy: 
- [ ] Check page source on any example (e.g. /deck.gl-raster/examples/cog-basic/) for the <script async src="https://plausible.io/js/pa-..."> tag
- [ ] Confirm pageviews appear in the Plausible dashboard.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -49,6 +49,12 @@ jobs:
             cp -r "$dir/dist/." "docs/static/examples/$example/"
           done
 
+      - name: Inject Plausible into examples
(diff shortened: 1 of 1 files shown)


kylebarron (approved):
I like this approach

- Automatically handles all examples without a manual approach
- Not something that clutters the sources of any example, so end users who might copy an example won't get plausible automatically

PR #484, opened 2026-04-27
fix: Allow `MinimalTileData` to be `null` in RasterTileLayer

kylebarron:
`null` is permitted to describe failed tile loads that do not produce any data, which then do not render any layer.


packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
```diff
@@ -24,18 +24,24 @@ import { TILE_SIZE, WEB_MERCATOR_TO_WORLD_SCALE } from "./constants.js";
...
-  width: number;
-  /**
-   * Byte length of the tile data, used by deck.gl's TileLayer for
-   * byte-based cache eviction when `maxCacheByteSize` is set. Optional.
-   */
-  byteLength?: number;
-};
+export type MinimalTileData =
+  | null
+  | undefined
```
kylebarron: eh probably nicer to only allow null not undefined

Issue #500, opened 2026-05-01
geotiff: multiplier for HTTP range sizes for header metadata


kylebarron:
ideally for the header we would make "multiplicative growing" header requests. If we fetch 16kb and that's not enough, then fetch 64kb, then fetch 1MB, and so on. I'm not sure where that fix would be, if that's here in our geotiff package or upstream in `@cogeotiff/core`.

----

### On multiplicative-growing reads

SourceChunk is stateless (@chunkd/middleware/build/src/middleware/chunk.js): it aligns each request to a fixed chunkSize. No memory of prior requests, no growth.

The natural place to add growing reads is `@chunkd/middleware` as a sibling middleware (e.g. `SourceGrowingChunk({ initial, max, factor })`):

- Initial reads use a small chunk
- Each subsequent fetch on the same source doubles the chunk up to max
- Drops in alongside SourceChunk/SourceCache in the existing pipeline

Implementing this in @cogeotiff/core doesn't fit — it consumes a Source interface but doesn't define source middleware. The defaultReadSize it exposes (now wired through your prefetch option) is only used for the very first byte read, then it falls back to range fetches sized to whatever it needs.

Two paths:

1. Upstream PR to @chunkd/middleware — best long-term, also benefits non-COG users.
2. Implement locally as a custom SourceMiddleware in our geotiff package, then upstream when stable. The SourceMiddleware interface is just fetch(req, next), easy to implement.

See https://github.com/developmentseed/async-tiff/pull/140

kylebarron:
In particular, we're currently always chunking our header reads into 1MB chunks, which is awful

kylebarron:
One other option is to always prefetch tags, but that seems wasteful and would lead to a slow first image byte

```ts
    // Pre-fetch tile layout for every non-primary IFD (overviews and masks).
    // `prefetchTags(primaryImage)` already covers the primary's TileByteCounts
    // and TileOffsets; without doing the equivalent for the rest, the first
    // `image.getTileSize(idx)` call on each overview triggers a chunked
    // `headerSource` read for those two tags. At default chunkSize that's two
    // 1MB requests per overview on first access, which dominates the network
    // panel and races abort signals during quick pans.
    const tileLayoutPrefetch = images.slice(1).flatMap((image) => [
      image.fetch(TiffTag.TileByteCounts, { signal }),
      image.fetch(TiffTag.TileOffsets, { signal }),
    ]);

    const [cachedTags] = await Promise.all([
[...]

PR #506, opened 2026-05-05
feat: Update land-cover example to have filterable categories

kylebarron:
https://github.com/user-attachments/assets/c7d70697-79b6-4d7a-baea-25c9bbf4453b

### Change list

- Changes `land-cover` example to upload the texture to the GPU as `uint8` non-normalized texture values
- Add boolean lookup table to decide from the selection UI which classification values should be drawn. Then, for any values where the lookup table resolves to `0`, discard those pixels.


kylebarron:
We _really_ need to implement https://github.com/developmentseed/deck.gl-raster/issues/500. Right now it's fetching 32kb regions at a time when reading header metadata

PR #509, opened 2026-05-06
feat(geotiff)!: exponential read-ahead cache for headers

### LINKED ISSUE #500: geotiff: multiplier for HTTP range sizes for header metadata

kylebarron:
This PR is being deprecated and closed in favor of #529, which should require much less total metadata bytes than this. In particular, that relaxes our existing requirement that we pre-fetch the entire TileByteCounts and TileByteOffsets arrays.

---

## Summary

- Replaces the fixed `[SourceChunk, SourceCache]` pipeline on `GeoTIFF.fromUrl` with a single sequential read-ahead cache that grows fetch sizes by a configurable multiplier — port of [async-tiff's `ReadaheadMetadataCache`](https://github.com/developmentseed/async-tiff/pull/140) to TypeScript.
- New internal `mutex()` helper (`packages/geotiff/src/concurrency.ts`) for serializing async tasks across awaits.
- New `SequentialBlockCache` + `SourceReadaheadCache` (`packages/geotiff/src/readahead-cache.ts`); the latter is a chunkd `SourceMiddleware`.
- **Breaking:** `GeoTIFF.fromUrl` drops `chunkSize` and `cacheSize` options. `prefetch` is kept (now meaning "initial fetch size") and a new `multiplier` option is added (defaults: 32 KiB / 2.0).

Closes #500

Spec: [`dev-docs/specs/2026-05-05-geotiff-readahead-cache-design.md`](https://github.com/developmentseed/deck.gl-raster/blob/kyle/geotiff-header-multiplier/dev-docs/specs/2026-05-05-geotiff-readahead-cache-design.md)

## Test Plan

- [x] `mutex()` unit tests — serialization, ordering, error isolation
- [x] `SequentialBlockCache` unit tests — single-block, cross-block, empty-buffer, EOF
- [x] `SourceReadaheadCache` unit tests — port of upstream's `test_readahead_cache`, plus bypasses (negative offset, full-file), concurrency serialization, file-size clamping, EOF break
- [x] Integration smoke test against `uint8_rgb_deflate_block64_cog` fixture — verifies the middleware integrates with the cogeotiff `Tiff` reader and triggers cache extension on small initial sizes
- [x] `pnpm --filter @developmentseed/geotiff typecheck` clean
- [x] All 86 tests in the package pass (the pre-existing `integration-rasterio.test.ts` failures are missing-fixture issues unrelated to this branch)


packages/geotiff/src/source/readahead-cache.ts
kylebarron, line 1: I think this should be inside a sub-folder of some name... maybe `cache`? Or `header`?

packages/geotiff/src/source/readahead-cache.ts
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * Intended to be called once `GeoTIFF.fromUrl` has finished its open-phase
+   * reads (`Tiff.create` + `prefetchTags(primaryImage)`). At that point the
+   * sequential read-ahead cache has done its job; subsequent reads from
+   * cogeotiff are at arbitrary offsets (lazy IFD lookups, GDAL ghost-header
+   * probes) and do not benefit from sequential-from-zero growth — in fact
+   * they cause catastrophic over-fetching as the cache grows exponentially
+   * to encompass each new far-offset request.
+   *
+   * Idempotent. One-way: there is no `enable()`.
```
kylebarron: I thought that we had separate sources for the header reads and the tile reads?

I think perhaps we need some more stable way to decide whether a tiff tag should be chunked with the start of the file or not. If there are other tags defined at the end of the file, and we hadn't called disable yet, then we'd still do a huge over fetch, right?

packages/geotiff/src/geotiff.ts
```diff
@@ -229,25 +229,42 @@ export class GeoTIFF {
...
    * @returns A Promise that resolves to a GeoTIFF instance.
    */
   static async fromUrl(
     url: string | URL,
     {
-      chunkSize = 1024 * 1024,
-      cacheSize = 10 * 1024 * 1024,
-      prefetch = 32 * 1024,
+      prefetch = 64 * 1024,
+      multiplier = 2,
```
kylebarron: I think perhaps this should be higher, like `4`, so we expand the fetch region faster

packages/geotiff/src/geotiff.ts
```diff
@@ -261,35 +278,36 @@ export class GeoTIFF {
...
       dataSource: source,
       headerSource: view,
       prefetch,
       signal,
     });
+
+    // Open phase complete: scope the cache to the open phase only. From here
+    // on, all reads (lazy overview tag lookups, GDAL ghost-header probes,
+    // etc.) bypass the cache and go straight to raw HTTP.
+    readahead.disable();
```
kylebarron: Perhaps we want a middle ground where we check the existing cache if the new range is already covered, but we also fetch at the end of the file separately if requested

packages/geotiff/src/overview.ts
```diff
@@ -118,9 +122,49 @@ export class Overview {
...
+   * everything except the primary image), this means many tiny per-tile
+   * range requests on every tile fetch. Calling
+   * `image.fetch(TiffTag.TileOffsets)` once forces cogeotiff to bulk-load
+   * the full array; thereafter all per-tile lookups are served from memory.
+   *
+   * The bulk fetch goes through the source originally passed to
+   * `Tiff.create` — for {@link GeoTIFF.fromUrl}, that's the wrapped header
+   * source. After {@link GeoTIFF.fromUrl} disables its read-ahead cache,
+   * the wrapper is a pass-through, so this read hits raw HTTP directly.
+   */
```
kylebarron, line 146: I'm not sure if we always want to fetch the entire tile offsets arrays for any input...

packages/geotiff/src/source/readahead-cache.ts
kylebarron: _(Written by Claude.)_

Done — moved to `packages/geotiff/src/source/` (and `packages/geotiff/tests/source/`) in 8d27c51. Went with `source/` rather than `cache/` or `header/` because the abstraction is a chunkd `SourceMiddleware` and `mutex` is a generic async helper, not a cache.

packages/geotiff/src/source/readahead-cache.ts
kylebarron: _(Written by Claude.)_

By Claude:

> Good catch — yes, this was a real risk. Addressed in 631f5b7 with a `maxExtension` cap on the cache (default 4 MiB). Now, even during the open phase (before `freeze()` is called), if satisfying a request would require pulling more than the cap into the cache, the middleware bypasses for that request and serves it with one direct fetch instead. So a GDAL ghost-header probe at EOF, or any other far-offset tag value, no longer triggers exponential growth.
>
> You're right that header and tile reads do use separate sources — `dataSource` (raw) for tile data via `fetchTile()`, `headerSource` (cached) for metadata. The hole was that cogeotiff's lazy per-tile offset/bytecount lookups also go through `headerSource`, which is why we needed both the cap and `freeze()` (now also via the lazy-prefetch in `Overview.ensureTagsLoaded`).

packages/geotiff/src/geotiff.ts
kylebarron: _(Written by Claude.)_

By Claude:

> Bumped to 4 in 631f5b7. The `maxExtension` cap (also new in that commit) makes this safe — even with the faster growth, no single fetch will exceed 4 MiB, so the worst case stays bounded for pathological files.

packages/geotiff/src/geotiff.ts
kylebarron: _(Written by Claude.)_

By claude:

> Done in 631f5b7 — renamed `disable()` to `freeze()` and changed its semantics to exactly that: cache hits are still served from memory, only misses bypass to the underlying source. A new `maxExtension` option (default 4 MiB) handles your other concern about far-offset reads — single requests that would require extending the cache by more than the cap bypass instead of triggering exponential growth, even while still active during the open phase.

packages/geotiff/src/overview.ts
kylebarron: By Claude:

> Fair concern. The trade-off: the alternative is per-entry lookups (each tile request triggers a 4–8 byte HTTP range for the offset and another for the bytecount), which scales linearly with the number of tiles fetched. For typical tile rendering — where many tiles from one overview are fetched in quick succession — the bulk fetch is one HTTP request and serves all subsequent lookups from memory.
>
> Array sizes are bounded: a 50K×50K image with 256×256 tiles has ~40K tiles → ~320 KB array; even a 100K×100K image is ~1.2 MB. We do this lazily per-overview, so users only pay for overviews they actually render from.
>
> If we want to make this opt-out (e.g. for very specific use cases like a single targeted tile fetch), happy to add a flag. Want me to file a follow-up issue, or do you have a specific case in mind where the bulk fetch is the wrong call?

packages/geotiff/src/source/readahead-cache.ts
```diff
@@ -92,8 +92,20 @@ export interface SourceReadaheadCacheOptions {
...
   multiplier: number;
+  /**
+   * Maximum bytes a single underlying fetch may add to the cache. If
+   * satisfying a request would require extending the cache by more than this,
+   * the middleware bypasses the cache entirely for that request and serves it
+   * with one direct fetch instead. Bounds the worst case when cogeotiff reads
+   * at a far offset during the open phase (e.g. GDAL ghost-header probes at
+   * EOF). Defaults to {@link DEFAULT_MAX_EXTENSION}.
+   */
+  maxExtension?: number;
```
kylebarron: Hmm, realistically how big is the largest header metadata we'll have?

kylebarron: This 200GB COG has a 61MB header. So I'm not sure if 4MB is enough for a per-fetch cache extension

https://vtopendata-prd.s3.amazonaws.com/Imagery/STATEWIDE_2025_30cm_LeafON_3Band.tif

packages/geotiff/src/geotiff.ts
```diff
@@ -234,16 +234,17 @@ export class GeoTIFF {
...
-   * is called: every subsequent fetch through the wrapped source becomes a
-   * pass-through to raw HTTP. This is intentional — cogeotiff/core lazily
-   * reads tile-offset/bytecount entries from the header source whenever a
-   * tile from a previously-untouched IFD is requested, and those reads are
-   * at arbitrary far offsets. With the cache active they would each pull
-   * the cache forward exponentially (e.g. a tile lookup at offset 8 MB with
-   * cache.len = 2 MB triggers a 4 MB underlying fetch). With it disabled,
-   * they go straight to raw HTTP and the cache stops mattering.
+   * The cache is **frozen at the end of the open phase**. Once `Tiff.create`
+   * and `prefetchTags(primaryImage)` finish, {@link SourceReadaheadCache.freeze}
```
kylebarron, line 238: Oh but I see we do always prefetch tile offsets and byte counts for the primary image already...

PR #511, opened 2026-05-07
fix(mosaic-layer): make sources prop reactive to updates

### LINKED ISSUE #510: Adding items to a mosaic layer
Right now, `MosaicLayer` only renders [sources](https://developmentseed.org/deck.gl-raster/api/deck-gl-geotiff/type-aliases/MosaicLayerProps/#sources) that are provided on initialization. Any later updates to `sources` are not propagated to the rendering.

This is motivated by STAC search in [stac-map](https://developmentseed.org/stac-map/), where I'd like to continuously update the map's mosaic as I load new items. My current workaround is to render one mosaic per page of search results:

https://github.com/user-attachments/assets/5ae78e85-6ef9-45bd-bc9b-2834663ec9de

Claude suggests:

> `TilesetClass` is only consumed once — when the inner `TileLayer` first initializes its `Tileset2D`. After that, deck.gl's `TileLayer` reuses the existing tileset across updates. So even though `MosaicLayer.renderLayers()` re-runs on every prop change and produces a new `TilesetClass` closing over the new `sources`, deck.gl never re-instantiates the tileset. Your new sources end up sitting in `props.sources` but the internal tileset is still the one built from your very first render. That matches what you're seeing: the layer doesn't pick up appended items.

kylebarron:
## Summary

- `MosaicTileset2D` now takes a `() => MosaicT[]` getter instead of a snapshot array; the spatial index rebuilds lazily inside `getTileIndices` whenever the closure returns a new array reference, picking up updates without recreating the tileset and preserving the inner tile cache.
- `MosaicLayer.renderTileLayer` closes over `this`, passing `() => self.props.sources` so deck.gl's existing `setOptions` → `_viewport = null` → `getTileIndices` flow on every prop update naturally surfaces the new sources.
- Public JSDoc on `MosaicLayerProps.sources` and `MosaicSource.x/y/z` now documents the array-reference-equality contract and the cache-stability implications for append vs. reorder/middle-removal.

Closes #510.

Design doc: [`dev-docs/specs/2026-05-07-mosaic-layer-dynamic-sources-design.md`](https://github.com/developmentseed/deck.gl-raster/blob/kyle/mosaic-layer-edit/dev-docs/specs/2026-05-07-mosaic-layer-dynamic-sources-design.md).

## Test plan

- [x] New `tests/mosaic-tileset.test.ts` covers initial query, append-without-reconstruction, empty → non-empty transition, in-place-mutation no-op (documents the reference-equality contract), default tile-id derivation from array position, explicit `x/y/z`, and zoom-range guards. 7/7 pass.
- [x] `pnpm exec biome check` clean on `packages/deck.gl-geotiff/src/mosaic-layer/` and the new test file.
- [x] `pnpm exec tsc --noEmit` reports no new errors in mosaic files (pre-existing workspace-build-order errors in `multi-cog-layer.ts` and `tests/{geotiff-tileset,render-pipeline}.test.ts` are unchanged).
- [ ] Smoke-test in `stac-map` against the use case in #510 (continuously appending STAC search results to a single `MosaicLayer`).

### DIFF
--- a/packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
+++ b/packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
@@ -83,13 +83,13 @@ export class MosaicLayer<
       maxRequests,
     } = this.props;
 
-    // Capture `this` so the factory's getter resolves the latest sources prop
-    // on every Tileset2D update cycle, allowing the spatial index to rebuild
-    // when the consumer passes a new sources array.
-    const self = this;
+    // The arrow function is defined here so its lexical `this` is the
+    // MosaicLayer instance (which deck.gl reuses across prop updates),
+    // ensuring `this.props.sources` returns the latest array on every call.
+    const getSources = () => this.props.sources;
     class MosaicTileset2DFactory extends MosaicTileset2D<MosaicT> {
       constructor(opts: any) {
-        super(() => self.props.sources, opts);
+        super(getSources, opts);
       }
     }
 
--- a/packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
+++ b/packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
@@ -18,7 +18,25 @@ export type MosaicLayerProps<
     | "maxCacheSize"
     | "maxRequests"
   > & {
-    /** List of mosaic sources to render */
+    /**
+     * List of mosaic sources to render.
+     *
+     * The mosaic updates reactively when this prop is replaced with a new
+     * array reference. Mutating the array in place will not trigger an
+     * update — pass a fresh array (e.g. `[...sources, newItem]`) to add or
+     * remove items.
+     *
+     * Tile cache reuse depends on stable tile IDs. By default, each source's
+     * tile ID is derived from its position in this array (see `MosaicSource`'s
+     * `x` / `y` / `z` for the exact derivation), so:
+     *
+     * - Appending items preserves all existing rendered tiles.
+     * - Reordering or removing items from the middle of the array invalidates
+     *   the cache slots of shifted items, causing them to re-fetch.
+     *
+     * Supply explicit `x`, `y`, and `z` identifiers per source if you need
+     * cache stability across arbitrary mutations of `sources`.
+     */
     sources: MosaicT[];
 
     /** Fetch data for this source. */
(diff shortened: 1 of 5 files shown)


packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
```diff
@@ -66,9 +83,13 @@ export class MosaicLayer<
       maxRequests,
     } = this.props;
 
+    // Capture `this` so the factory's getter resolves the latest sources prop
+    // on every Tileset2D update cycle, allowing the spatial index to rebuild
+    // when the consumer passes a new sources array.
+    const self = this;
```
kylebarron: hmm do we need to explicitly redefine `this`?

packages/deck.gl-geotiff/src/mosaic-layer/mosaic-tileset-2d.ts
```diff
@@ -28,32 +36,59 @@ export type MosaicSource = {
...
+ * closure returns a new array reference (compared by `===`); mutating the
+ * array in place will not be detected.
  */
 export class MosaicTileset2D<MosaicT extends MosaicSource> extends Tileset2D {
-  private sources: (TileIndex & MosaicT)[];
-  private index: Flatbush;
+  private getSources: () => MosaicT[];
+  private cachedRaw: MosaicT[] | null = null;
+  private sources: (TileIndex & MosaicT)[] = [];
+  private index: Flatbush | null = null;
```
kylebarron: this could benefit from a comment on each

packages/deck.gl-geotiff/src/mosaic-layer/mosaic-tileset-2d.ts
```diff
@@ -66,15 +101,20 @@ export class MosaicTileset2D<MosaicT extends MosaicSource> extends Tileset2D {
...
     if (viewport.zoom > (maxZoom ?? Infinity)) {
       return [];
     }
+    if (this.cachedRaw === null || this.cachedRaw.length === 0) {
+      return [];
+    }
 
     const bounds = viewport.getBounds();
-    const indices = this.index.search(...bounds);
+    const indices = this.index!.search(...bounds);
```
kylebarron: Instead of using `!` I feel like `this.ensureIndex` should return the index...

kylebarron (commented):
This LGTM! I'll wait for @gadomski to check this against stac-map

kylebarron:
I think this overall looks good; we can merge and revisit in case there's any unexpected behavior

PR #518, opened 2026-05-08
fix: traverse tiles across world copies (#517)

### LINKED ISSUE #517: Rendering in multiple world copies
@alukach brought up that we have some issues in rendering data across multiple world copies.

In particular, based on this example in cog-basic, it's **not** that we don't render across multiple world copies (you can see that working when zoomed fully out), it's that our viewport tile traversal only considers the viewport of the primary world. This is why, when zooming in just on the right side of the antimeridian, it slowly begins to remove tiles just on the left side of the antimeridian... as those same tiles are no longer visible on the _right side_ of the screen. 

https://github.com/user-attachments/assets/2e762087-92d5-4cbe-b5a4-2ea44b60cb1c

When panning across the antimeridian left to right, you see the raster jump between the primary viewport as it switches back and forth.

https://github.com/user-attachments/assets/f39bd577-84eb-4fd6-aadf-0026ac0c6592

This isn't an issue with the antimeridian per se, because the data itself does not _cross_ the antimeridian.

kylebarron:
https://github.com/user-attachments/assets/7dfbd215-135e-4b0d-ad3e-24c225aae5f0

------

## Summary

- Re-runs `RasterTileNode.update()` at world offsets ±1, ±2, ... ±`MAX_MAPS` when `viewport.subViewports?.length > 1`, so tiles drawn in adjacent world copies pass frustum culling instead of getting evicted.
- Threads `worldOffset` through `getBoundingVolume`; non-zero offsets return a translated copy derived from the offset-0 cache.
- Makes non-zero passes purely additive — they may flip `selected` from `false` → `true` but never the reverse — to avoid LOD-flip pathologies when the OBB center moves with the offset.
- Bounds check always uses the offset-0 AABB (the dataset bounds live in primary-world common space; the tile represents the same data at every offset).

Mirrors the upstream pattern in [`@deck.gl/geo-layers/tile-2d-traversal.ts`](https://github.com/visgl/deck.gl/blob/b0134f025148b52b91320d16768ab5d14a745328/modules/geo-layers/src/tileset-2d/tile-2d-traversal.ts).

Closes #517.

## Design + docs

- Spec: [`dev-docs/specs/2026-05-07-multi-world-copy-tile-traversal-design.md`](dev-docs/specs/2026-05-07-multi-world-copy-tile-traversal-design.md)
- Living doc: [`dev-docs/world-copies.md`](dev-docs/world-copies.md)

## Test Plan

- [x] `pnpm --filter @developmentseed/deck.gl-raster test --run` — 93 passing (was 86; +7 multi-world tests).
- [x] `pnpm typecheck` — clean.
- [ ] Manual verification in `cog-basic`: pan east of +180° / west of −180° at a moderate zoom; tiles should remain visible across the antimeridian instead of disappearing as in the issue's videos.

## Out of scope

- The large-zarr-root cull path in `createRootTiles` does not yet consider world-copy intersections. Tracked as a follow-up if a zarr dataset surfaces the same symptom.


kylebarron:
This looks good but I don't have the mental energy to review this right now

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -380,19 +422,43 @@ export class RasterTileNode {
...
     zRange: ZRange,
     project: ((xyz: number[]) => number[]) | null,
     boundingVolumeCache: BoundingVolumeCache,
+    worldOffset = 0,
   ): { boundingVolume: OrientedBoundingBox; commonSpaceBounds: Bounds } {
     const hit = boundingVolumeCache.get(this.z, this.x, this.y);
+    let base: {
+      boundingVolume: OrientedBoundingBox;
+      commonSpaceBounds: Bounds;
+    };
```
kylebarron, line 457: Maybe just a one-line comment here that says base is in the primary world offset, and will be translated if necessary

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -380,19 +422,43 @@ export class RasterTileNode {
...
+   *                             cached offset-0 volume directly. Non-zero
+   *                             values return a fresh translated copy.
    */
   getBoundingVolume(
     zRange: ZRange,
     project: ((xyz: number[]) => number[]) | null,
     boundingVolumeCache: BoundingVolumeCache,
+    worldOffset = 0,
   ): { boundingVolume: OrientedBoundingBox; commonSpaceBounds: Bounds } {
     const hit = boundingVolumeCache.get(this.z, this.x, this.y);
```
kylebarron: Nit: rename to `cacheHit` for clarity

PR #525, opened 2026-05-11
fix(raster-tileset): memoize tile bounding volumes across traversals

### LINKED ISSUE #523: Tile traversal regression: getTileIndices ~12 ms/frame on HiDPI (no cross-frame memoization)
I noticed a slowdown on tile traversal while testing all the examples during #521 

Here `getTileIndices` takes nearly 100% of CPU for the usgs-topo-cutline example. I think this example is especially pronounced because its tiles are so small (128px x 128px)

Image

---

## Summary

`getTileIndices` (raster tile traversal) takes ~12 ms per animation frame on a HiDPI display, causing dropped frames. Visible in the `usgs-topo-cutline` example (and other examples) during the `fitBounds` transition or any pan/zoom.

This is a regression from #513 / 4f3e5d0e8bf4b34a3481b235580d926ea81caf0c ("Use `devicePixelRatio` for level of detail computation in tile traversal") — but the `devicePixelRatio` change is correct; it merely exposed an existing inefficiency in the traversal.

## Mechanism

1. #513 changed the LOD test from "one source pixel ≤ one **CSS** pixel" to "≤ one **device** pixel" ([`raster-tile-traversal.ts`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L295-L315)) and the layer wires `getPixelRatio` to `drawingBufferWidth / cssWidth` ([`raster-tile-layer.ts`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts#L274-L296)). On a 2× display the threshold halves, so the traversal descends **one extra overview level over the whole viewport** → ~4× more tiles selected (this is by design, closes #64).

2. `getTileIndices` runs **once per animation frame** while the viewport is changing, and it is **stateless across frames**:
   - [`createRootTiles`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L615-L663) allocates fresh `RasterTileNode`s every call.
   - Every visited node calls [`getBoundingVolume` → `_getGenericBoundingVolume`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L423-L479), which does **9 proj4 forward reprojections** ([`REF_POINTS_9`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L80-L86)) plus a `makeOrientedBoundingBoxFromPoints` (covariance + eigendecomposition), recomputed from scratch.
   - The per-node `_boundingVolume` / `_children` caches are never reused because the nodes themselves are discarded each frame.

So per-frame cost ≈ `O(visited nodes) × (9 × proj4.forward + 1 × OBB)`. The COG path uses proj4's generic `forward()` pipeline ([`cog-layer.ts`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-geotiff/src/cog-layer.ts#L199-L231)), which is the heavy bit being run 9× per node. ~4× the node count → ~3 ms → ~12 ms → dropped frames.

## Repro

`usgs-topo-cutline` example on a Retina display: every animation frame spends ~12 ms in `getTileIndices` (visible as partially dropped frames in the Chrome DevTools performance panel during the load `fitBounds` transition / panning).

A synthetic EPSG:4326 COG descriptor (USGS-quad-shaped, ~zoom 11 viewport) through `getTileIndices` confirms the level shift:

| `pixelRatio` | level selected | tiles selected | wall time/call (synthetic) |
|---|---|---|---|
| 1 | z=2 | 4 | 0.13 ms |
| 2 | z=3 | 25 | 0.29 ms |

(The synthetic wall-time ratio understates the real case — it uses a hand-rolled mercator forward, ~10–50× cheaper than proj4's generic `forward()`.)

## Fix directions

The device-pixel change is fine; the real problem is that frame-invariant work (tree structure + bounding volumes) is recomputed every frame. Roughly in order of payoff:

1. **Memoize the traversal node tree on the `RasterTileset2D` across `getTileIndices` calls** (root-cause fix). Keep root nodes (and their lazily-built `_children` / `_boundingVolume`) alive between frames, with a count-bounded cache (analogous to deck.gl's `maxCacheSize`). Per-frame cost collapses to frustum culling + the scalar LOD compare; proj4/OBB run once per tile. Makes the device-pixel change essentially free per frame and is robust regardless of traversal depth.
2. **Axis-aligned fast path for EPSG:4326 / mercator-family sources** — the "Case 2/3" TODOs already in [`getBoundingVolume`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L383-L416). A lat/lng tile is an axis-aligned box in EPSG:3857 — 2 corner reprojections + an AABB instead of 9 points + an oriented box. ~4–5× cheaper per node; covers all the USGS examples. Complementary to #1.
3. **`REF_POINTS_9` → `REF_POINTS_5`** for LOD/culling sampling — halves proj4 calls per node, small quality tradeoff.
4. **Policy: don't default the LOD criterion to full device pixels** — make it opt-in, or clamp (e.g. `min(dpr, ~1.5)`), or expose a layer prop. Avoids the 4× tile fetch *and* the 4× traversal but gives back some of the #64 sharpness win.

Minor aside: `makeClampedForwardTo3857` is applied in [`cog-layer.ts`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-geotiff/src/cog-layer.ts#L209-L231) and *again* inside [`sampleReferencePointsInEPSG3857`](https://github.com/developmentseed/deck.gl-raster/blob/728244fd399a7d886fbc8fdda7d0ea8fabc58c55/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts#L538-L564) — harmless but redundant.

## Related

- #513 (introduced the device-pixel LOD criterion), #64 (motivation)
- #516 (split tile CPU/GPU caching) — related caching theme

kylebarron:
Closes #523 

This is a whole lot better than in #523. In particular we have almost no dropped frames in the usgs topo cutline example:

image

There's still more perf improvements we could do in the future, but this is still a big improvement.

-------------

## What I am changing

Fixes the per-frame tile-traversal regression in #523. On a HiDPI display, `getTileIndices` (the raster tile traversal) was taking ~12 ms/frame — visible as dropped frames in the `usgs-topo-cutline` example during the load `fitBounds` transition / panning. `getTileIndices` runs once per animation frame and was recomputing every visited tile's bounding volume from scratch (≈9 proj4 reprojections + an oriented-bounding-box fit per tile), because nothing survived between calls. #513's switch to a device-pixel LOD criterion — correct, and kept — made this ~4× worse on 2× displays by descending one extra overview level over the viewport.

## How I did it

Memoize bounding volumes across `getTileIndices` calls. A tile's bounding volume depends only on `(z, x, y, zRange)` for a given tileset descriptor — frame-invariant — so it's safe to cache.

- **`BoundingVolumeCache`** (new, internal): an LRU `Map` keyed `"z/x/y"`, soft-capped (default 65 536 entries), `sweep()`ed once at the top of each traversal so a single frame is never starved of an entry it computed earlier that frame. It caches only *visited* tiles, populated lazily — for a single COG that's ~10³ entries total; the cap exists to bound huge single-level zarrs / long pan sessions.
- **`RasterTileset2D`** owns one cache for its lifetime (so it persists across animation frames) and threads it into the traversal. New `RasterTileset2DOptions.maxBoundingVolumeCacheSize` tunes the cap; no new layer prop. Standalone callers of `getTileIndices` that don't pass a cache get a throwaway one (still dedups within a traversal).
- The cache flows through `update()`'s params (alongside the viewport / frustum / etc., which already propagate down the recursion), and `getBoundingVolume(zRange, project, cache)` is now just a cache wrapper around a lower-level `computeBoundingVolume` that holds the case dispatch (Globe assert, future mercator fast-path slots, generic Case 4). The old per-node `_boundingVolume` field is removed — it never actually hit (`update` calls `getBoundingVolume` once per node).

Steady-state per-frame traversal cost drops from `O(visited nodes) × (≈9 proj4 + OBB fit)` to `O(visited nodes) × (Map lookup + frustum cull + LOD compare)` — well under a millisecond. No behavior change: the selected tile set for a given viewport / `zRange` / `pixelRatio` is identical.

Design doc: [`dev-docs/specs/2026-05-11-traversal-bounding-volume-cache-design.md`](dev-docs/specs/2026-05-11-traversal-bounding-volume-cache-design.md).

## How you can test it

- `pnpm --filter @developmentseed/deck.gl-raster test` — 93 tests pass, including new `bounding-volume-cache.test.ts` (LRU / sweep behavior), `bounding-volume-cache-memoization.test.ts` (a `projectTo3857`-counting descriptor proves a second `getTileIndices` call does zero reprojections), and `raster-tileset-2d-cache.test.ts` (cache reuse + correctness with a tiny / zero cap).
- Manually: open the `usgs-topo-cutline` example on a HiDPI display and profile in the Chrome DevTools performance panel — per-frame `getTileIndices` is now sub-millisecond after the first frame, with no dropped frames during the load `fitBounds` transition / panning.

## Related Issues

Fixes #523.

Out of scope, tracked in #523 as follow-ups: an axis-aligned fast path in `computeBoundingVolume` for EPSG:4326 / mercator-family sources (cheaper *cold* compute), and removing the redundant double-wrap of `makeClampedForwardTo3857`.

### DIFF
--- a/packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
+++ b/packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
@@ -24,12 +24,12 @@ export interface BoundingVolumeCacheOptions {
    * so a single frame is never starved of an entry it computed earlier that
    * same frame. `0` makes every traversal start from an empty cache.
    *
-   * @default 65536
+   * @default 65_536
    */
   maxEntries?: number;
 }
 
-const DEFAULT_MAX_ENTRIES = 1 << 16;
+const DEFAULT_MAX_ENTRIES = 65_536;
 
 /**
  * An LRU cache of tile bounding volumes keyed by `"z/x/y"`.
--- a/packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
+++ b/packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
@@ -1,0 +1,108 @@
+import type { OrientedBoundingBox } from "@math.gl/culling";
+
+import type { Bounds, ZRange } from "./types.js";
+
+/**
+ * A memoized tile bounding volume, tagged with the elevation range it was
+ * computed for (so a `zRange` change can invalidate it).
+ */
+export interface BoundingVolumeCacheEntry {
+  zRange: ZRange;
+  boundingVolume: OrientedBoundingBox;
+  commonSpaceBounds: Bounds;
+}
+
+/**
+ * Options for {@link BoundingVolumeCache}.
+ */
+export interface BoundingVolumeCacheOptions {
+  /**
+   * Soft cap on the number of cached tile bounding volumes. When a
+   * `getTileIndices` traversal begins and the cache holds more than this many
+   * entries, the least-recently-used entries are dropped down to roughly half.
+   * Eviction never runs mid-traversal (only via {@link BoundingVolumeCache.sweep}),
+   * so a single frame is never starved of an entry it computed earlier that
+   * same frame. `0` makes every traversal start from an empty cache.
+   *
+   * @default 65536
+   */
+  maxEntries?: number;
+}
+
+const DEFAULT_MAX_ENTRIES = 1 << 16;
+
+/**
+ * An LRU cache of tile bounding volumes keyed by `"z/x/y"`.
+ *
+ * The raster tile traversal recomputes a tile's bounding volume (several proj4
+ * reprojections plus an oriented-bounding-box fit) only on a cache miss; on a
+ * hit it returns the stored volume. A tile's bounding volume depends only on
+ * `(z, x, y, zRange)` for a given tileset descriptor, so it is safe to memoize
+ * across `getTileIndices` calls (i.e. across animation frames).
+ *
+ * NOTE: this assumes a non-Globe traversal (`project === null`) — the only kind
+ * `RasterTileNode.getBoundingVolume` currently supports. If Globe-view bounding
+ * volumes are ever implemented, the cache key will need a viewport/resolution
+ * component.
+ */
+export class BoundingVolumeCache {
+  private entries = new Map<string, BoundingVolumeCacheEntry>();
+  private maxEntries: number;
+
+  constructor({
+    maxEntries = DEFAULT_MAX_ENTRIES,
+  }: BoundingVolumeCacheOptions = {}) {
+    this.maxEntries = Math.max(0, maxEntries);
+  }
+
+  /** Number of cached entries. */
+  get size(): number {
+    return this.entries.size;
+  }
+
+  /**
+   * Look up the cached bounding volume for tile `(z, x, y)`. On a hit the entry
+   * is marked most-recently-used. Returns `undefined` on a miss.
+   */
+  get(z: number, x: number, y: number): BoundingVolumeCacheEntry | undefined {
+    const key = `${z}/${x}/${y}`;
+    const entry = this.entries.get(key);
+    if (entry === undefined) {
+      return undefined;
+    }
+    // Re-insert to move the key to the most-recently-used end of the Map.
+    this.entries.delete(key);
+    this.entries.set(key, entry);
+    return entry;
+  }
+
+  /** Store the bounding volume for tile `(z, x, y)` as most-recently-used. */
+  set(z: number, x: number, y: number, entry: BoundingVolumeCacheEntry): void {
+    const key = `${z}/${x}/${y}`;
+    this.entries.delete(key);
+    this.entries.set(key, entry);
+  }
+
+  /**
+   * If the cache is over its soft cap, drop least-recently-used entries down to
+   * roughly half of `maxEntries`. No-op when at or under the cap. Call once at
+   * the start of a traversal, never mid-traversal.
+   */
+  sweep(): void {
+    if (this.entries.size <= this.maxEntries) {
+      return;
+    }
+    const target = Math.floor(this.maxEntries / 2);
+    const excess = this.entries.size - target;
+    const keysToDelete: string[] = [];
+    for (const key of this.entries.keys()) {
+      keysToDelete.push(key);
+      if (keysToDelete.length >= excess) {
+        break;
+      }
+    }
+    for (const key of keysToDelete) {
+      this.entries.delete(key);
+    }
+  }
+}
--- a/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
+++ b/packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
@@ -28,7 +28,7 @@ import {
 } from "@math.gl/culling";
 import { lngLatToWorld, worldToLngLat } from "@math.gl/web-mercator";
 
-import type { BoundingVolumeCache } from "./bounding-volume-cache.js";
+import { BoundingVolumeCache } from "./bounding-volume-cache.js";
 import type { TilesetDescriptor, TilesetLevel } from "./tileset-interface.js";
 import type {
   Bounds,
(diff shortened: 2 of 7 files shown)


packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Eviction never runs mid-traversal (only via {@link BoundingVolumeCache.sweep}),
+   * so a single frame is never starved of an entry it computed earlier that
+   * same frame. `0` makes every traversal start from an empty cache.
+   *
+   * @default 65536
+   */
+  maxEntries?: number;
+}
+
+const DEFAULT_MAX_ENTRIES = 1 << 16;
```
kylebarron: It's more readable to just define this as `65_536`

packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * A memoized tile bounding volume, tagged with the elevation range it was
+ * computed for (so a `zRange` change can invalidate it).
+ */
+export interface BoundingVolumeCacheEntry {
+  zRange: ZRange;
+  boundingVolume: OrientedBoundingBox;
+  commonSpaceBounds: Bounds;
+}
```
kylebarron, line 13: Isn't ZRange a property of the data? Well, I guess looking at the [`TileRange.zRange` docs](https://deck.gl/docs/api-reference/geo-layers/tile-layer#zrange), it _is_ a property of the data but it's not necessarily known by the individual tiles?

packages/deck.gl-raster/src/raster-tileset/bounding-volume-cache.ts
```diff
@@ -0,0 +1,108 @@
...
+   */
+  get(z: number, x: number, y: number): BoundingVolumeCacheEntry | undefined {
+    const key = `${z}/${x}/${y}`;
+    const entry = this.entries.get(key);
+    if (entry === undefined) {
+      return undefined;
+    }
+    // Re-insert to move the key to the most-recently-used end of the Map.
+    this.entries.delete(key);
+    this.entries.set(key, entry);
```
kylebarron, line 75: Is deleting and re-inserting the easiest way to handle the LRU order?

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -153,16 +154,31 @@ export class RasterTileNode {
     result: { boundingVolume: OrientedBoundingBox; commonSpaceBounds: Bounds };
   };
 
+  /**
+   * Optional cross-traversal bounding-volume cache, shared by every node in a
+   * traversal. When set, {@link getBoundingVolume} reads/writes it instead of
+   * the per-node `_boundingVolume` field, so bounding volumes survive across
+   * `getTileIndices` calls. Provided by `RasterTileset2D`.
+   */
+  private boundingVolumeCache?: BoundingVolumeCache;
```
kylebarron: Why optional? What's the downside to making it always exist? We can default to a newly-created cache with default parameters if one isn't passed in.

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -200,9 +216,15 @@ export class RasterTileNode {
...
       const { descriptor } = this;
+      const boundingVolumeCache = this.boundingVolumeCache;
       for (let y = minRow; y <= maxRow; y++) {
         for (let x = minCol; x <= maxCol; x++) {
-          children.push(new RasterTileNode(x, y, childZ, { descriptor }));
+          children.push(
+            new RasterTileNode(x, y, childZ, {
+              descriptor,
+              boundingVolumeCache,
+            }),
```
kylebarron: Is this a little weird? Feels like the cache is a property of the node, rather than the nodes being a property of the cache?

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -384,13 +406,27 @@ export class RasterTileNode {
...
-    ) {
-      return cached.result;
+    const boundingVolumeCache = this.boundingVolumeCache;
+
+    // Cache lookup. A tile's bounding volume depends only on (z, x, y, zRange)
+    // for a given descriptor, so the shared cache (when the tileset provides
+    // one) is reused across getTileIndices calls / animation frames. Without a
+    // shared cache we fall back to the per-node `_boundingVolume` field (good
+    // for one traversal only), so standalone callers behave exactly as before.
+    if (boundingVolumeCache) {
```
kylebarron: If the bounding volume cache always exists, then we can simplify this if/else block

packages/deck.gl-raster/src/raster-tileset/raster-tile-traversal.ts
```diff
@@ -399,34 +376,22 @@ export class RasterTileNode {
...
-      if (
-        cached &&
-        cached.zRange[0] === zRange[0] &&
-        cached.zRange[1] === zRange[1]
-      ) {
-        return cached.result;
-      }
+    const hit = boundingVolumeCache.get(this.z, this.x, this.y);
+    if (hit && hit.zRange[0] === zRange[0] && hit.zRange[1] === zRange[1]) {
+      return hit;
```
kylebarron, line 391: Perhaps we should split `getBoundingVolume` and then a lower level `computeBoundingVolume`? `getBoundingVolume` checks the cache and returns if a hit was found. Otherwise it calls `computeBoundingVolume` and caches the result before returning it.

Issue #524, opened 2026-05-11
SourceError: Request outside of bounds


kylebarron:
The land-cover example currently fails on `main`.

<img width="2168" height="1421" alt="Image" src="https://github.com/user-attachments/assets/d60d4eed-766b-4803-b8a3-6a26d5b017ec" />

Curiously, this is the _only_ example that currently fails with this error.

Same as https://github.com/developmentseed/stac-map/issues/459 / https://github.com/developmentseed/stac-map/pull/460 probably. cc @gadomski 

These are the erroring requests:

```
tile-layer.js:20 SourceError: Request outside of bounds bytes=1727583-1727586
    at SourceChunk.fetch (chunk.js:27:19)
    at view.js:44:40
    at SourceView.run (view.js:48:16)
    at SourceView.fetch (view.js:36:21)
    at TiffImage.getTileSize (tiff.image.js:521:50)
    at async getTile (fetch.ts:347:33)
    at async fetchCogBytes (fetch.ts:242:20)
    at async Promise.all (land-cover/index 0)
    at async fetchTile (fetch.ts:59:34)
    at async Overview.fetchTile (overview.ts:121:12)

tile-layer.js:20 SourceError: Request outside of bounds bytes=1727872-1727875
    at SourceChunk.fetch (chunk.js:27:19)
    at view.js:44:40
    at SourceView.run (view.js:48:16)
    at SourceView.fetch (view.js:36:21)
    at TiffImage.getTileSize (tiff.image.js:521:50)
    at async getTile (fetch.ts:347:33)
    at async fetchCogBytes (fetch.ts:242:20)
    at async Promise.all (land-cover/index 0)
    at async fetchTile (fetch.ts:59:34)
    at async Overview.fetchTile (overview.ts:121:12)
tile-layer.js:20 SourceError: Request outside of bounds bytes=1291400-1291403
    at SourceChunk.fetch (chunk.js:27:19)
    at view.js:44:40
    at SourceView.run (view.js:48:16)
    at SourceView.fetch (view.js:36:21)
    at TiffImage.getTileSize (tiff.image.js:521:50)
    at async getTile (fetch.ts:347:33)
    at async fetchCogBytes (fetch.ts:242:20)
    at async Promise.all (land-cover/index 0)
    at async fetchTile (fetch.ts:59:34)
    at async Overview.fetchTile (overview.ts:121:12)
tile-layer.js:20 SourceError: Request outside of bounds bytes=1556670-1556673
[...]

PR #529, opened 2026-05-12
perf(geotiff)!: block-aligned LRU header cache; lazy tile metadata

### LINKED ISSUE #294: Figure out good caching and chunking defaults for `GeoTIFF.fromUrl`
Ref https://github.com/blacha/cogeotiff/issues/1431

### LINKED ISSUE #500: geotiff: multiplier for HTTP range sizes for header metadata

### LINKED ISSUE #501: geotiff: Helper to get byte length of header
This is a claude attempt I don't like:

It works by iterating through the tile offsets and picking the smallest one...

Instead, we should implement this as a middleware injected (it could be auto-injected) to the GeoTIFF load. Then we can just record the byte ranges that were fetched?

But that also doesn't work because it'll give us a range _past_ the end of the header metadata.

We really need to hook into the underlying tag parsing I think...

We should create an issue for this upstream

```ts
  /**
   * The byte length of the TIFF header region — i.e. the offset at which
   * tile data begins, measured from the start of the file.
   *
   * Computed as the minimum of `TileOffsets` across every IFD (primary +
   * overviews + masks). For COGs with `LAYOUT=IFDS_BEFORE_DATA` (which
   * GDAL-produced COGs use) this equals the total bytes occupied by all
   * IFD chains and their tag arrays.
   *
   * Useful for tuning `chunkSize` / `prefetch` in {@link GeoTIFF.fromUrl}:
   * if `headerByteLength` is larger than the configured prefetch, the
   * header walk will require multiple range requests.
   *
   * Async because `TileOffsets` is not in cogeotiff's `ImportantTags` set
   * loaded eagerly during `init`, so overview/mask offsets must be fetched
   * on demand. The fetches are cached by the underlying source so a second
   * call is effectively free.
   */
  async getHeaderByteLength(): Promise<number> {
    const allImages: TiffImage[] = [
      this.image,
      ...(this.maskImage ? [this.maskImage] : []),
    ];
    for (const ov of this.overviews) {
      allImages.push(ov.image);
      if (ov.maskImage) {
        allImages.push(ov.maskImage);
      }
    }
    const offsetArrays = await Promise.all(
      allImages.map((image) => image.fetch(TiffTag.TileOffsets)),
    );
    let min = Number.POSITIVE_INFINITY;
    for (const offsets of offsetArrays) {
      if (!offsets) {
        continue;
      }
      for (const o of offsets as number[]) {
        const n = Number(o);
        if (n > 0 && n < min) {
          min = n;
        }
      }
    }
    return Number.isFinite(min) ? min : 0;
  }
```

### LINKED ISSUE #528: geotiff: Reduce latency for visualizing large COGs
I think I want a wholesale reconsideration of how we load metadata from COGs. We want to balance the twin goals of: 

- low latency for large files
- reducing the total number of header requests, and reducing the latency to load each additional image tile. That is, in a perfect world where it didn't cost anything to load the header metadata, we would load everything up front because it would mean that each additional image tile could be loaded in one single request. 

But right now for really large COGs like in the vermont cog comparison example, we're fetching 60MB of metadata before we can render any single image tile. In that case it would be much better to avoid pre-fetching the entire tile offsets and byte counts. Especially because the initial view might not be at the highest resolution anyways, so the full-resolution image's offsets and byte counts might not be used for the initial render anyways.

kylebarron:
This provides a huge latency improvement for rendering large COGs.

### Before

61MB of header fetching!

https://github.com/user-attachments/assets/f2ae27bd-0a4c-4b0a-8870-462e7c6eb94d

### After:

Roughly 3 seconds from a cold cache, and much less metadata. (4x more tiles are being fetched in this view than in the Before, because https://github.com/developmentseed/deck.gl-raster/pull/513 increased tile count by 4x)

https://github.com/user-attachments/assets/61808530-969c-4dc2-8981-2eb2f6673c72

### Change list

- Don't prefetch big tags for tile byte counts and tile offsets
- Use block-aligned cache only for header metadata, split from tile requests
- Add `debug?` param to `GeoTIFF.open` for logging out debug info about requests made

Closes https://github.com/developmentseed/deck.gl-raster/issues/528, Closes https://github.com/developmentseed/deck.gl-raster/issues/501, Closes https://github.com/developmentseed/deck.gl-raster/issues/294

---------

## Summary

Reworks COG header loading after live testing showed PR #509's approach (eager bulk-prefetch of TileOffsets/TileByteCounts at open) downloaded tens of MB before any tile could render on huge COGs.

This branch goes in the opposite direction — lazy everything, with a fixed-block LRU cache to amortize per-entry reads. Matches geotiff.js's `BlockedSource` architecture.

- **Drop eager `TileOffsets`/`TileByteCounts` prefetch.** cogeotiff already supports lazy per-entry reads; the block cache (below) makes them cheap.
- **Disable cogeotiff's GDAL leader-bytes path** in `GeoTIFF.open` (`tiff.options = undefined`). That optimization assumes tiles fit in one chunk and is harmful when they don't — it pollutes the header cache with image-data bytes.
- **`GeoTIFF.fromUrl` uses 64 KiB `SourceChunk` + 8 MiB `SourceCache`** (matches geotiff.js's defaults). LRU-ish eviction keeps memory bounded.
- **Breaking:** `fromUrl` drops the `prefetch` option (still on `GeoTIFF.open` for direct callers).

For the Vermont 200 GB COG with 61 MB header: previous PR downloaded ~60 MB upfront; this design opens with ~3 small reads and lazy-loads per-IFD metadata as tiles are actually requested.

Closes #500. Supersedes #509 (which should be closed).

Spec: [`dev-docs/specs/2026-05-12-cog-block-cache-design.md`](https://github.com/developmentseed/deck.gl-raster/blob/kyle/cog-block-cache/dev-docs/specs/2026-05-12-cog-block-cache-design.md)

## Test plan

- [x] `pnpm --filter @developmentseed/geotiff typecheck` clean
- [x] Full vitest suite passes (69/69, excluding pre-existing `integration-rasterio` fixture failures unrelated to this branch)
- [x] New `block-cache.test.ts` asserts: every underlying fetch is 64 KiB-aligned; `tiff.options` is undefined post-open; `getTileSize(0)` triggers at most 2 chunk fetches
- [x] Existing `fromurl.test.ts` (issue #524 regression) still passes
- [ ] Manual: open the Vermont 200 GB COG in the example, confirm browser network panel shows small initial reads (~few hundred KB total) instead of tens of MB
- [ ] Manual: zoom around the Vermont COG, confirm tile rendering works and per-tile metadata reads are served from cached blocks after warmup


dev-docs/specs/2026-05-12-cog-block-cache-design.md
```diff
@@ -0,0 +1,128 @@
...
+**Status:** Design — supersedes [`2026-05-05-geotiff-readahead-cache-design.md`](2026-05-05-geotiff-readahead-cache-design.md) (and the unmerged PR [#509](https://github.com/developmentseed/deck.gl-raster/pull/509)).
+
+## Background
+
+The original design (the sequential exponential read-ahead cache, then a frozen-after-open variant) tried to optimize *steady-state* tile rendering by bulk-loading `TileOffsets` / `TileByteCounts` arrays for each IFD. That moved the cost to *open time*. On a real 200 GB Vermont COG, that's tens of MB downloaded before any tile renders — even though the initial view is at an overview level whose primary-image arrays are never used.
+
+geotiff.js takes the opposite approach. Each `fromUrl` call fetches just 1024 bytes (header + first IFD pointer). `getImage(i)` reads only that IFD's entries; tile-array values are wrapped in a `DeferredArray` that holds only their file offset + count. Per-tile reads fetch a single 4–8 byte entry through `BlockedSource`, a fixed-block LRU that coalesces adjacent entries into one block. The block cache lives inside the source layer; cogeotiff's lazy per-entry reads benefit from it automatically.
```
kylebarron: We should fetch an initial prefetch of a full block size, not 1024 bytes

dev-docs/specs/2026-05-12-cog-block-cache-design.md
```diff
@@ -0,0 +1,128 @@
...
+- No tunable that has to be right per file. Pathological cases (huge metadata regions, far-offset probes) all degrade gracefully — they just cost more block fetches.
+
+### Why LRU eviction?
+
+The previous design's sequential cache *never evicted*. For long-running sessions or large files, that's a memory leak. `SourceCache` is a two-generation cache (cacheA flips to cacheB on overflow, cacheB drops) — not strict LRU but bounded and approximately recency-aware in practice.
+
+### Why disable cogeotiff's leader-bytes path?
+
+cogeotiff auto-detects the GDAL ghost option `BLOCK_LEADER=SIZE_AS_UINT4` at `Tiff.create()` time. If present, `TiffImage.getTileSize()` skips the `TileByteCounts` lookup and instead fetches 4 bytes just before the tile data. The comment in cogeotiff explains the intent: *"This fetch will generally load in the bytes needed for the image too provided the image size is less than the size of a chunk."* But that assumption breaks for tiles larger than the block size (very common — many COG tiles are 256×256×3 bytes ≈ 200 KB, well above 64 KiB). When it breaks, the result is:
```
kylebarron, line 52: How does it know where the tile data is though? Does it fetch the offset separately?

packages/geotiff/src/geotiff.ts
```diff
@@ -111,6 +111,14 @@ export class GeoTIFF {
...
       signal,
     });
+    // Disable cogeotiff's GDAL leader-bytes path so `TiffImage.getTileSize`
+    // always reads from TileOffsets/TileByteCounts through the header source.
+    // The leader-bytes optimization assumes a tile fits in one chunk, which
+    // breaks for typical 256x256x3 tiles (~200 KB) vs. our 64 KiB blocks.
+    // Without this, the leader read pulls image-data bytes into the header
+    // cache, evicting metadata. cogeotiff core only reads `tiff.options` in
+    // that one path, so nulling it here is safe.
+    tiff.options = undefined;
```
kylebarron, line 139: It would be nicer to have a cleaner way to turn off the GDAL leader-bytes path. I'll create an upstream issue

packages/geotiff/src/geotiff.ts
```diff
@@ -269,33 +272,23 @@ export class GeoTIFF {
...
-    // read files in chunks
-    const chunk = new SourceChunk({ size: chunkSize });
-    // 10MB cache for recently accessed chunks
-    const cache = new SourceCache({ size: cacheSize });
-
-    const view = new SourceView(source, [chunk, cache]);
+    const view = new SourceView(source, [
+      new SourceChunk({ size: chunkSize }),
+      new SourceCache({ size: cacheSize }),
+    ]);
```
kylebarron, line 309: Are we sure we're using the right source for header vs data loads?

For example 
https://github.com/developmentseed/deck.gl-raster/blob/201ff27618d88bd5bbf6cfe7af1f2f0e44927eab/packages/geotiff/src/fetch.ts#L313-L319
we should probably rename `source` to be clearer about `dataSource` and `headerSource`

packages/geotiff/src/geotiff.ts
kylebarron: https://github.com/blacha/cogeotiff/issues/1467

packages/geotiff/src/overview.ts
```diff
@@ -70,6 +70,11 @@ export class Overview {
     return this.geotiff.nodata;
   }
 
+  /** Inherits the {@link GeoTIFF.debug} flag from the parent. */
+  get debug(): boolean {
+    return this.geotiff.debug;
+  }
+
```
kylebarron: I don't think we need to make this public

packages/geotiff/src/geotiff.ts
```diff
@@ -96,22 +107,36 @@ export class GeoTIFF {
...
+    debug?: boolean;
   }): Promise<GeoTIFF> {
-    const { dataSource, headerSource, prefetch = 32 * 1024, signal } = options;
+    const { dataSource, headerSource, signal, debug } = options;
+    // We use cogeotiff's default read size; in the typical fromUrl path,
+    // SourceChunk pads any small request up to the block size anyway, so
+    // tuning this independently of the chunk size is rarely useful.
     const tiff = await Tiff.create(headerSource, {
-      defaultReadSize: prefetch,
+      defaultReadSize: Tiff.DefaultReadSize,
```
kylebarron: Can't we just omit this? And it'll use `Tiff.DefaultReadSize` by default?

Issue #537, opened 2026-05-13
Failing publish of 0.7.0-beta.1


kylebarron:
Trying to publish 0.7.0-beta.1 [failed](https://github.com/developmentseed/deck.gl-raster/actions/runs/25821422344/job/75863747993) with 

```
npm notice Publishing to https://registry.npmjs.org/ with tag beta and default access
npm notice publish Signed provenance statement with source and build information from GitHub Actions
npm notice publish Provenance statement published to transparency log: https://search.sigstore.dev/?logIndex=1525512661
npm error code E400
npm error 400 Bad Request - PUT https://registry.npmjs.org/@developmentseed%2faffine - OIDC publish authorize: Invalid token
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2026-05-13T19_29_09_586Z-debug-0.log
```
not sure why.

Also failed after https://github.com/developmentseed/deck.gl-raster/pull/535

kylebarron:
Fixed by #538

PR #538, opened 2026-05-13
ci: Explicitly unset NODE_AUTH_TOKEN during publishing

kylebarron:
An attempt to solve #537. Ref https://github.com/actions/setup-node/issues/1440#issuecomment-3571890875


kylebarron:
This worked

PR #546, opened 2026-05-14
feat: Expose onTile callbacks through raster tile layers

kylebarron:
Add more props to RasterTileLayer (and thus also COGLayer and ZarrLayer)

- onTileLoad
- onTileUnload
- onTileError
- onViewportLoad

Also add props to MosaicLayer:

- onSourceLoad
- onSourceUnload
- onSourceError
- onViewportLoad

In the case of the MosaicLayer, it's just an implementation detail that it's a TileLayer under the hood, so we use the term `source` consistently.


kylebarron:
cc @yharby

yharby:
Thank you @kylebarron

Issue #273, opened 2026-02-27
Extend deck.gl tile fetching to enable multi-tile fetches


kylebarron:
**Problem statement**: currently, deck.gl has just one available callback for fetching tiles: `TileLayer.getTileData`. This is user-provided and is always called with _just one_ tile at a time. This means that _all fetches_ are necessarily independent. 

For most historical uses of the deck.gl TileLayer this is not a problem. For sources like MVT, the tile responses usually come from different URLs in a single domain, and there would be no possibility to coalesce requests. 

**However** for newer data sources like PMTiles or COG, each individual tile data source comes from the _same URL_. This means that we're unnecessarily making many requests when we could potentially make many fewer by coalescing tile ranges. E.g. if we have 20 tiles in the current viewport, we currently are forced to always make 20 requests, whereas if it's a grid of 5 columns by 4 rows, then with request coalescing we could make no more than 4 requests (one per row, where each row's bytes are adjacent in memory)

**Requirements**: 

- [ ] Update `Source` in `@cogeotiff/core` to have a `fetchRanges` option, with a `maxCoalesceBytes`
- [ ] Update `Tiff` in `@cogeotiff/core` to have a `fetchTiles` option, which will fetch multiple tiles at once, coalescing up to `maxCoalesceBytes`
- [ ] Update deck.gl `TileLayer` with a new user-facing prop `getTileDataBatched` that takes in an array of tiles and fetches all of them. This should be optional, falling back to `getTileData` if it doesn't exist.
- [ ] Update `Tileset2D` class to call `getTileDataBatched` if it exists. Ref https://github.com/visgl/deck.gl/blob/32c5d8d91dea69bc14839283c5bbc2600603725b/modules/geo-layers/src/tileset-2d/tileset-2d.ts#L264
- [ ] Update `COGLayer` to pass in `getTileDataBatched`, which calls the underlying `cogeotiff`'s `fetchTiles`

kylebarron:
I made an upstream PR to `@cogeotiff/core` to support fetching multiple tiles at once, with range coalescing: https://github.com/blacha/cogeotiff/pull/1463

Also tracked in https://github.com/developmentseed/deck.gl-raster/issues/407

PR #557, opened 2026-05-19
feat: per-origin HTTP concurrency limiter for COG layers

### LINKED ISSUE #555: MosaicLayer: Connect RequestScheduler of the top-level layer with all sub layers
As it is, we have request schedulers that are fighting each other inside of each individual TileLayer instance.

kylebarron:
**Huge reduction in network requests with the MosaicLayer**

**before,** scrolling around generates many many requests that aren't cancelled when panning around the map. You can see we fetch almost **300MB** of data **before** loading the data in the end viewport location

https://github.com/user-attachments/assets/5c3f7d7b-0d9b-41c1-85de-42e02676bbdb

**after** we've only fetched **150MB** of data _even after a ton of scrolling_. **The request cancellation is working much better**:

https://github.com/user-attachments/assets/fe9b4bce-6948-46d6-a783-dbac06ce31fd

Closes https://github.com/developmentseed/deck.gl-raster/issues/555

---- 

## Summary

Adds a shared, per-origin HTTP concurrency limiter that caps concurrent tile-data fetches per host across all `COGLayer` / `MultiCOGLayer` / `MosaicLayer` instances on a page, with dynamic, viewport-aware priority so center-of-screen sources jump the queue.

**The problem this addresses.** Most COGs we target live on AWS S3 (or similar), which serves HTTP/1.1 only. Browsers cap concurrent HTTP/1.1 connections per origin at ~6 (Chrome). deck.gl's `Tileset2D` ships a `RequestScheduler({ maxRequests: 6 })` *per-`TileLayer` instance* — so two `COGLayer`s on the same S3 bucket can each schedule 6 → ≥12 in flight → the browser queues the excess. Worse, those queued requests stick around when the viewport pans, blocking fresh requests behind stale ones. The cap has to be **per origin, shared across layers**, and the queue has to **re-order on pan** so newly-central sources are serviced before stale edge sources.

## What's in this PR

**In `@developmentseed/geotiff`:**

- **`ConcurrencyLimiter` interface** — `acquire(url, signal?, getPriority?: () => Priority): Promise<() => void>`. Signal-aware: a queued request whose signal aborts (e.g. user panned away) is dropped before any network I/O fires, no slot burned.
- **`Priority` type** — `number | readonly number[]`. Numbers compare numerically; arrays compare lexicographically with missing trailing elements treated as 0, so `[1]` and `[1, 0, 0]` are equivalent and `[1, -1]` jumps ahead of `[1]`. Lower = serviced sooner.
- **Dynamic priority** — `getPriority` is a callback, re-invoked **on every slot-open**. Panning the viewport re-sorts the queue without any explicit "reprioritize" call. A linear-scan find-min on release is the right primitive here (a heap can't represent priorities that move under it).
- **`PerOriginSemaphore` class** — default impl. Maintains `Map<origin, Semaphore>` internally; one `PerOriginSemaphore({ maxRequests: 6 })` shared across layers gates each origin to 6 in flight, independently.
- **`LimiterMiddleware` class** — chunkd `SourceMiddleware`, shape matches `SourceChunk` / `SourceCache`. Slotted into the existing header-source `SourceView` after the cache, so cache hits short-circuit and never burn a slot — only uncached network reads gate. Data source path gets a tiny `SourceView` of its own with just the middleware, so every tile-data fetch gates.
- **`GeoTIFF.fromUrl(url, { concurrencyLimiter?, getPriority? })`** — accepts a limiter (non-null → installed; `null` or omitted → no gating) and an optional priority callback forwarded to the limiter for every fetch through this GeoTIFF's sources.
- `Semaphore` (the FIFO + priority primitive) stays internal. Public surface: `ConcurrencyLimiter` (type), `Priority` (type), `PerOriginSemaphore` (class).

**In `@developmentseed/deck.gl-geotiff`:**

- **`defaultConcurrencyLimiter`** — a module-level `new PerOriginSemaphore({ maxRequests: 6 })`. Referenced from `COGLayer`, `MultiCOGLayer`, and `MosaicLayer`'s `defaultProps.concurrencyLimiter`, so any two layers on the same origin share one cap **out of the box, zero user setup**.
- **`concurrencyLimiter?: ConcurrencyLimiter | null`** prop on `COGLayer`, `MultiCOGLayer`, and `MosaicLayer`. Pass your own to override; pass `null` to disable.
- **`MosaicLayer` priority** — computes a `getPriority` per source (euclidean distance from `source.bbox` center to the current viewport center, in degree-space) and threads it, along with the layer's `concurrencyLimiter`, through `getSource`'s `opts`. A consumer's `getSource` can spread `opts` straight into `GeoTIFF.fromUrl` so both wire up automatically.
- `fetchGeoTIFF` accepts the options and forwards them to `GeoTIFF.fromUrl`.

## Cancellation flow

1. deck.gl pans → `Tileset2D._pruneRequests` aborts unselected in-flight tiles.
2. Each tile's `AbortSignal` propagates through `getTileData` → `fetchTile` → `dataSource.fetch(..., { signal })`.
3. `LimiterMiddleware` forwards the signal to `limiter.acquire(url, signal, getPriority)`:
   - Already aborted → reject immediately, no slot consumed.
   - Aborted while queued in the inner `Semaphore` → splice from queue, reject, no slot consumed.
   - Aborted in-flight (after acquire) → underlying `fetch` aborts via its own signal handling; `finally` releases the slot.

## Priority flow

1. `MosaicLayer.getTileData` builds a `getPriority` closure that reads `this.context.viewport` lazily, so each invocation sees the current viewport (not the one at queue time).
2. The closure is passed to `getSource(opts)`; example app spreads `opts` into `GeoTIFF.fromUrl`.
3. `LimiterMiddleware` forwards it to every `acquire` call.
4. `Semaphore.release` linear-scans pending waiters, calls each one's `getPriority()`, picks the min, and resumes it. Panning the viewport between releases mutates the values the next scan sees — no explicit re-queue.

## Out of scope (deferred)

- **Multi-tile request coalescing** (the `TileBatcher` / `getMultiTileData` thread from the design discussion). The interaction with deck.gl's `_pruneRequests` (which only fires when `ongoing > maxRequests`) made it more involved than expected — gating + priority first is cleanly mergeable and unblocks the cross-layer pain point; coalescing follows.
- Subclassing `Tileset2D` to fire abort signals on `onTileUnload` regardless of `maxRequests`.
- Upstream deck.gl proposals (`getTileDataBatched`, public `pruneRequests`, etc.) — captured in the spec.
- A standalone `@developmentseed/concurrency` package — the limiter primitives live in `@developmentseed/geotiff` for now; extractable later when/if a non-geotiff source format wants to share an instance.

## Spec

[`dev-docs/specs/2026-05-19-concurrency-limiter-design.md`](dev-docs/specs/2026-05-19-concurrency-limiter-design.md).

## Test plan

- [x] `pnpm --filter @developmentseed/geotiff test` — passes. New: `Semaphore` (FIFO + signal + dynamic priority) tests, `PerOriginSemaphore` tests, `LimiterMiddleware` tests, `fromUrl({ concurrencyLimiter, getPriority })` integration tests with `SourceHttp.fetch` stubbed. The 17 pre-existing `integration-rasterio.test.ts` failures are unrelated (missing local `.npy` fixtures).
- [x] `pnpm --filter @developmentseed/deck.gl-geotiff test` — passes. `concurrency-limiter.test.ts` confirms `COGLayer` / `MultiCOGLayer` / `MosaicLayer` all default `concurrencyLimiter` to the same module-level `PerOriginSemaphore` instance.
- [x] `pnpm --filter @developmentseed/geotiff typecheck` + `pnpm --filter @developmentseed/deck.gl-geotiff typecheck` — clean.
- [x] `pnpm biome check packages/geotiff packages/deck.gl-geotiff examples/naip-mosaic` — clean.
- [x] Manual: `naip-mosaic` example with ~100 COGs in view — first rendered tile is now near viewport center (was bottom-left), time-to-first-render dropped from ~5s to sub-second.

Part of #273 (the multi-tile coalescing other half is a follow-up).


packages/geotiff/src/limiter.ts
```diff
@@ -0,0 +1,76 @@
+/** A pending acquire waiting for a slot. */
+interface Waiter {
+  resolve(release: () => void): void;
+  reject(reason: unknown): void;
+  signal?: AbortSignal;
+  onAbort?: () => void;
```
kylebarron, line 57: I think a small docstring on each of these would be useful

packages/geotiff/src/limiter.ts
```diff
@@ -0,0 +1,76 @@
...
+          }
+        };
+        waiter.onAbort = onAbort;
+        signal.addEventListener("abort", onAbort, { once: true });
+      }
+      this.queue.push(waiter);
+    });
+  }
+
+  private _makeRelease(): () => void {
```
kylebarron, line 125: Small docstring would be useful.

packages/geotiff/src/limiter.ts
```diff
@@ -0,0 +1,76 @@
...
+    return () => {
+      if (released) {
+        return;
+      }
+      released = true;
+      this._releaseOne();
+    };
+  }
+
+  private _releaseOne(): void {
```
kylebarron, line 140: Small docstring would be useful.

packages/geotiff/src/limiter.ts
```diff
@@ -74,3 +74,47 @@ export class Semaphore {
     next.resolve(this._makeRelease());
   }
 }
+
+/**
+ * Minimal contract for capping concurrent {@link Source.fetch} calls. An
+ * implementation hands out slots scoped however it likes; the default
+ * {@link PerOriginSemaphore} scopes per `url.origin`.
+ */
```
kylebarron, line 171: Just double checking: there's no way for JS to see whether requests are being made over HTTP/1 vs HTTP/2 or 3, right?

packages/geotiff/src/geotiff.ts
```diff
@@ -308,9 +313,21 @@ export class GeoTIFF {
...
+    // Wrap the *raw* data source's `.fetch` when a limiter is supplied so
+    // every tile-data fetch holds a slot for its duration. Header reads go
+    // through the cached SourceView and are not gated.
+    const dataSource: Pick<Source, "fetch"> = concurrencyLimiter
+      ? {
+          fetch: limitFetch(
+            source.fetch.bind(source),
+            new URL(url),
+            concurrencyLimiter,
+          ),
```
kylebarron: I think it would be cleaner if we defined a middleware class similar to `SourceChunk` and `SourceCache` that we could slot in, rather than overriding the fetch like this.

Also what's the benefit of not limiting header fetches? Those still count too, right?

examples/naip-mosaic/src/App.tsx
```diff
@@ -85,10 +85,18 @@ type TextureDataT = {
  */
 const geotiffCache = new Map<string, Promise<GeoTIFF>>();
 
-function getCachedGeoTIFF(url: string, signal?: AbortSignal): Promise<GeoTIFF> {
+function getCachedGeoTIFF(
+  url: string,
+  opts: Parameters<typeof GeoTIFF.fromUrl>[1],
```
kylebarron: I don't like this as a type hint; we should have a type alias for this; maybe GeoTIFFUrlOptions? Or GeoTIFFCreationOptions?

examples/naip-mosaic/src/App.tsx
```diff
@@ -85,10 +85,18 @@ type TextureDataT = {
...
+): Promise<GeoTIFF> {
   let promise = geotiffCache.get(url);
   if (!promise) {
-    promise = GeoTIFF.fromUrl(url, { signal }).catch((err) => {
+    // MosaicLayer threads its `concurrencyLimiter` and `getPriority` through
+    // `opts`, so spreading here forwards both to `GeoTIFF.fromUrl` — every
+    // header + tile-data fetch through this GeoTIFF's sources is gated
+    // against the shared per-origin pool and re-ordered dynamically by
+    // distance from viewport center.
```
kylebarron: I don't think we need to have this verbose comment

packages/deck.gl-geotiff/src/geotiff/geotiff.ts
```diff
@@ -54,9 +58,23 @@ export function addAlphaChannel(rgbImage: RasterArray): RasterArray {
...
+     *  `concurrencyLimiter` is set. */
+    getPriority?: () => Priority;
+  } = {},
 ): Promise<GeoTIFF> {
   if (typeof input === "string" || input instanceof URL) {
-    return await GeoTIFF.fromUrl(input);
+    return await GeoTIFF.fromUrl(input, {
+      concurrencyLimiter: options.concurrencyLimiter,
+      getPriority: options.getPriority,
+    });
```
kylebarron: Seems like this fetchGeoTIFF should pass through a `signal` too? From the top-level COGLayer prop?

packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
```diff
@@ -190,9 +224,43 @@ export class MosaicLayer<
...
+          // Viewport exposes longitude/latitude for geographic viewports;
+          // fall back to (0, 0) defensively if not available.
+          const lon =
+            "longitude" in viewport ? (viewport.longitude as number) : 0;
+          const lat =
+            "latitude" in viewport ? (viewport.latitude as number) : 0;
+          const dx = sourceCx - lon;
+          const dy = sourceCy - lat;
+          return Math.hypot(dx, dy);
+        };
```
kylebarron: Instead of this, I feel like we should explicitly check for WebMercatorViewport or GlobeViewport, and if the viewport isn't an instance of one of those, then we don't define `getPriority` at all?

packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
```diff
@@ -190,9 +224,43 @@ export class MosaicLayer<
...
+        // deck.gl fills `concurrencyLimiter` from `defaultProps` when the
+        // user doesn't supply it; an explicit `null` is the opt-out signal
+        // and gets forwarded as-is so `GeoTIFF.fromUrl` skips installing
+        // the middleware. Only `undefined` (e.g. types-only escape hatches)
+        // falls back to the default here.
+        const concurrencyLimiter =
+          this.props.concurrencyLimiter === undefined
+            ? defaultConcurrencyLimiter
+            : this.props.concurrencyLimiter;
```
kylebarron: I think here `defaultConcurrencyLimiter` will have already been set on `this.props.concurrencyLimiter` by `defaultProps`, right?

packages/deck.gl-geotiff/src/cog-layer.ts
```diff
@@ -189,7 +208,9 @@ export class COGLayer<
   }
 
   async _parseGeoTIFF(): Promise<void> {
-    const geotiff = await fetchGeoTIFF(this.props.geotiff);
+    const geotiff = await fetchGeoTIFF(this.props.geotiff, {
+      concurrencyLimiter: this.props.concurrencyLimiter,
+    });
```
kylebarron: Shouldn't we pass down an abort signal too?

packages/deck.gl-geotiff/src/default-concurrency-limiter.ts
```diff
@@ -0,0 +1,12 @@
+import { PerOriginSemaphore } from "@developmentseed/geotiff";
+
+/**
+ * Shared default concurrency limiter for every COGLayer / MultiCOGLayer that
+ * doesn't override its `concurrencyLimiter` prop. A single module-level
+ * `PerOriginSemaphore({ maxRequests: 6 })` so two layers fetching from the
+ * same origin (e.g. the same S3 bucket) share *one* HTTP/1.1 connection
+ * pool. The cap matches Chrome's default per-origin HTTP/1.1 limit.
+ */
+export const defaultConcurrencyLimiter = new PerOriginSemaphore({
```
kylebarron: I feel that the global default should be in UPPER_CASE_CONSTANT

packages/deck.gl-geotiff/src/multi-cog-layer.ts
```diff
@@ -334,7 +347,9 @@ export class MultiCOGLayer extends RasterTileLayer<
     // Open all COGs in parallel
     const cogSources = await Promise.all(
       entries.map(async ([name, config]) => {
-        const geotiff = await fetchGeoTIFF(config.url);
+        const geotiff = await fetchGeoTIFF(config.url, {
+          concurrencyLimiter: this.props.concurrencyLimiter,
+        });
```
kylebarron: Similarly, we should be passing in an abort signal here right?

packages/deck.gl-geotiff/src/cog-layer.ts
```diff
@@ -168,20 +168,28 @@ export class COGLayer<
...
   } as typeof RasterTileLayer.defaultProps;
 
   declare state: {
     geotiff?: GeoTIFF;
     tilesetDescriptor?: RasterTilesetDescriptor;
     defaultGetTileData?: COGLayerProps<TextureDataT>["getTileData"];
     defaultRenderTile?: COGLayerProps<TextureDataT>["renderTile"];
+    /** Aborts the in-flight header read when the `geotiff` prop changes or the
+     *  layer is removed, freeing its limiter slot for fresh work. */
+    abortController?: AbortController;
```
kylebarron, line 183: I'm not sure we need to store our own abortController here; that feels like overkill. What does it give us? I don't think the geotiff prop is going to be changing often here. I think it matters more for the MosaicLayer, where its sources might be changing, but it'll be creating _different_ COGLayer instances for different COG urls

kylebarron: Hmm; but it means that creating and then removing this layer will stop its header reads? Perhaps that's worth this complexity

packages/geotiff/src/fetch.ts
```diff
@@ -326,14 +326,15 @@ async function findBandSeparateTileByteRanges(
   self: HasTiffReference,
   x: number,
   y: number,
+  signal?: AbortSignal,
```
kylebarron: I think this should be in an options object so it doesn't have to be wrapped and unwrapped from an object reference. I think that's simpler and also faster, right?

packages/deck.gl-geotiff/src/cog-layer.ts
kylebarron: If any layer would need the AbortController, wouldn't it be the MosaicLayer?

packages/deck.gl-geotiff/src/mosaic-layer/mosaic-layer.ts
```diff
@@ -190,9 +233,43 @@ export class MosaicLayer<
...
+          ? (): number => {
+              // Re-read the viewport each call so panning re-sorts the queue.
+              const viewport = this.context.viewport as
+                | WebMercatorViewport
+                | _GlobeViewport;
+              const dx = sourceCx - viewport.longitude;
+              const dy = sourceCy - viewport.latitude;
+              return Math.hypot(dx, dy);
+            }
+          : undefined;
```
kylebarron: Can we put this into a helper function instead of having it inline? It's verbose.

Maybe something like `function createGetPriorityCallback`?

packages/deck.gl-geotiff/src/cog-layer.ts
kylebarron: From claude:

> no — it opens GeoTIFFs in getSource during a tile load, where deck.gl already hands it a per-tile signal. The controller belongs in COGLayer/MultiCOGLayer precisely because they open in updateState with no built-in signal.

packages/deck.gl-geotiff/src/multi-cog-layer.ts
```diff
@@ -331,18 +356,36 @@ export class MultiCOGLayer extends RasterTileLayer<
...
+          return { name, geotiff, sourceProjection };
+        }),
+      );
+    } catch (err) {
+      // Layer removed mid-open (finalizeState aborted the signal); drop it.
+      if (signal?.aborted) {
+        return;
+      }
+      throw err;
+    }
```
kylebarron: I think this might be clearer as a helper function

kylebarron:
Ok after a LOT of back and forth I think this PR is ready to merge!

Issue #512, opened 2026-05-07
Loss of precision (jitter) with high resolution imagery


kylebarron:
At high resolutions (imagery of at least ~1m/pixel), we get a "jitter" with the imagery moving around as we pan in/out.

This is due to imprecision of only using float32 instead of float64 on the GPU.

Here's an example in the naip-mosaic example:

https://github.com/user-attachments/assets/6ea93c39-0cf6-43ab-9ae7-d99e5b54f455

I know upstream deck.gl has a smart workaround for this, where the change the georeferencing model above a certain zoom, though I can't remember exactly how it works. We should probably adopt that same model.

PR #558, opened 2026-05-19
Spike: high-zoom precision via per-tile reference point (superseded by #559)

kylebarron:
This branch: lots of jitter, and in particular **per tile independent jitter** so we have little gaps between tiles. Jitter noticeably starts above ~z16

https://github.com/user-attachments/assets/cbcd2e56-648e-4642-a067-4fd7f828d1a2

-----

> **Closed without merging — this is a record of an approach to #512 that did not work.** The working fix is **#559**. See that PR (and `dev-docs/coordinate-systems.md` on it) for the full analysis.

## What this tried

Attack the high-zoom jitter (#512) by keeping mesh vertex magnitudes small: store each vertex as an **offset from a per-tile reference point** (the tile centroid in EPSG:3857 meters) instead of an absolute coordinate, then add the reference back via the layer transform. Iterations on this branch (and follow-on experiments in the session that produced it):

1. Offset subtraction in the mesh + per-tile reference baked into `modelMatrix.translation`.
2. Per-tile `coordinateOrigin` (let deck.gl's auto-offset subtract it).
3. Shared per-frame `coordinateOrigin = viewport.center`, per-tile shift in `modelMatrix.translation`.

## Why it didn't work

- **Tile seams.** Encoding a *shared boundary vertex* differently in adjacent tiles (e.g. `+305` in tile A, `−305` in tile B) makes the two tiles run different float32 arithmetic chains for the same physical point. The sub-pixel disagreement gets amplified by GPU rasterization coverage rounding into visible 1-px seams — which `main` does not have (it stores the same absolute value in both tiles, so they're bit-identical). The seam is structural to *any* per-tile encoding, regardless of where the reference shift lives.
- **No real precision win.** Even when numerically sub-pixel, the auto-offset origin uniform (`shaderCoordinateOrigin`) is itself float32 at ~10² world-unit magnitude, leaving a ~1 m floor — indistinguishable from `main`.

## What actually fixed it (#559)

Render the mesh in **common space** with an **identity model matrix** and **`coordinateOrigin = [0,0,0]`** (so `shaderCoordinateOrigin` is exactly `Math.fround(viewport.center)` and the camera-relative subtraction is Sterbenz-exact), plus **fp64 attribute pairs** for the vertex positions. Absolute common-space coords keep shared boundary vertices bit-identical → no seams. Full write-up in #559.

Related: #512

### DIFF
--- a/packages/deck.gl-raster/src/raster-layer.ts
+++ b/packages/deck.gl-raster/src/raster-layer.ts
@@ -106,6 +106,20 @@ export interface RasterLayerProps extends CompositeLayerProps {
 
   /** Opacity of the debug overlay. */
   debugOpacity?: number;
+
+  /**
+   * Reference point in projected (EPSG:3857) coordinates. When provided,
+   * mesh vertex positions are stored in the GPU buffer as float32 offsets
+   * from this point, rather than as absolute EPSG:3857 meters. Used by
+   * {@link RasterTileLayer} to preserve precision at high zoom levels.
+   *
+   * Callers passing this prop must fold the same reference point back into
+   * their `modelMatrix` translation (in float64 on the CPU) so the
+   * algebraic output is identical to the no-reference case.
+   *
+   * @default null
+   */
+  referencePointMeters?: [number, number] | null;
 }
 
 const defaultProps: DefaultProps<RasterLayerProps> = {
@@ -113,6 +127,7 @@ const defaultProps: DefaultProps<RasterLayerProps> = {
   // deck.gl (as long as async: true)
   image: { type: "image", value: null, async: true },
   renderPipeline: { type: "array", value: [], compare: true },
+  referencePointMeters: { type: "array", value: null, compare: true },
   debug: false,
   debugOpacity: 0.5,
 };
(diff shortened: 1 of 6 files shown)


packages/deck.gl-raster/src/raster-tileset/raster-tileset-2d.ts
```diff
@@ -48,6 +49,17 @@ export type RasterTileMetadata = {
    */
   projectedCorners: Corners;
 
+  /**
+   * Reference point in projected (EPSG:3857) coordinates, used by
+   * {@link RasterLayer} to split mesh vertex positions into a float32 offset
+   * and a float64 origin to preserve precision at high zoom. Computed once
+   * per tile as the centroid of {@link RasterTileMetadata.projectedCorners}.
+   *
+   * See `dev-docs/specs/2026-05-19-high-zoom-precision-design.md` for the
```
kylebarron, line 58: Before merging the PR, I think this is core enough to warrant a top-level dev-doc explainer. Perhaps it should be something like "projection handling" explaining not just high zoom precision handling, and in the future could be expanded with globe view information.

packages/deck.gl-raster/src/raster-tileset/raster-tileset-2d.ts
```diff
@@ -48,6 +49,17 @@ export type RasterTileMetadata = {
...
+   * reasoning.
+   */
+  referencePointMeters: Point;
```
kylebarron, line 61: Are we sure this is always going to be in meters?

packages/deck.gl-raster/src/raster-layer.ts
```diff
@@ -170,12 +185,19 @@ export class RasterLayer extends CompositeLayer<RasterLayerProps> {
       props.reprojectionFns.inverseReproject !==
         oldProps.reprojectionFns?.inverseReproject;
 
+    const refChanged =
```
kylebarron, line 188: spell out `referencePointChanged`

kylebarron:
Written by claude

> Closing as superseded by #559. Keeping the branch and this PR as a record of the per-tile-reference approach that produced tile seams and didn't beat main's precision floor — see the PR description and #559 for the analysis.

PR #559, opened 2026-05-19
fix: Use fp64 emulation to fix lack of precision (jitter) at high zooms

### LINKED ISSUE #512: Loss of precision (jitter) with high resolution imagery

kylebarron:
**After**: smooth, high-precision zoom in/out without any jitter!

https://github.com/user-attachments/assets/a9c3485f-b7ce-4ef2-9720-ce2da9c290b7

**Before**:

https://github.com/user-attachments/assets/37648f51-1384-497f-89e6-7a276706f483

https://github.com/user-attachments/assets/6ea93c39-0cf6-43ab-9ae7-d99e5b54f455

-----

## Summary

Fixes #512 — raster tiles jittered and drifted off the basemap above ~z16 over high-resolution imagery (sub-meter NAIP).

**Root cause:** mesh vertex positions in EPSG:3857 meters (~10⁷ magnitude) quantize to ~1 m when written into a float32 GPU attribute. Two further float32-at-large-magnitude steps in deck.gl's cartesian + auto-offset shader chain compound it (the scale multiply, and the auto-offset origin re-quantization).

**Fix** — eliminate all three float32 error sources:

- **Render the mesh in deck.gl common space** (world units) instead of 3857 meters: `forwardReproject`/`inverseReproject` are wrapped to apply `meters × scale + TILE_SIZE/2` (memoized per-descriptor for reference stability), so `modelMatrix` is identity (kills the scale-multiply rounding) and `coordinateOrigin` is `[0,0,0]` (so `shaderCoordinateOrigin` is exactly `Math.fround(viewport.center)` → no upload re-quantization → Sterbenz-exact camera-relative subtraction).
- **fp64 attribute pairs** for vertex positions (high + low Float32Arrays, via the shader's existing `position64Low` path) recover the attribute quantization.

Shared boundary vertices stay **bit-identical** across adjacent tiles (absolute common-space coords + deterministic split), so there are no tile seams. Delatin's triangulation is unchanged: its error metric is in pixel space, and the affine wrap commutes with barycentric interpolation.

A dev-mode assertion guards the `SimpleMeshLayer` per-instance-transform invariants the fp64 correction relies on. The full error analysis and deck.gl precision background are documented in `dev-docs/coordinate-systems.md` and `dev-docs/specs/2026-05-19-high-zoom-precision-design.md`.

## Previous attempt

#558 records the per-tile-reference-point approach that didn't work (tile seams + no real precision gain over `main`) — closed and kept for reference.

## Test plan

- [ ] Unit tests pass (`pnpm --filter @developmentseed/deck.gl-raster test`) — fp64 split reconstruction (<1e-6 m), invariant assertion
- [ ] NAIP mosaic example at z16–z19: pan/zoom shows no jitter relative to basemap, no seams between tiles (verified ✅)
- [ ] Low zoom (z4–z11) rendering unchanged
- [ ] No new GL/console warnings


dev-docs/coordinate-systems.md
```diff
@@ -0,0 +1,347 @@
+# Coordinate systems and precision in deck.gl
+
+> Source references in this doc target
+> [`visgl/deck.gl@82a02831`](https://github.com/visgl/deck.gl/tree/82a028314b8b20275c8f58713e68702407f2eba4)
+> (deck.gl 9.3.x). Function/line numbers may drift; the prose explains the
+> mechanism so you can re-locate it if so.
```
kylebarron: We should use permalinks to this commit, including specific lines

packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
```diff
@@ -403,32 +447,45 @@ export class RasterTileLayer<
...
+      deckProjectionProps = {
+        coordinateSystem: "cartesian",
+        coordinateOrigin: [0, 0, 0],
+        // biome-ignore format: array
+        modelMatrix: [
+          1, 0, 0, 0,
+          0, 1, 0, 0,
+          0, 0, 1, 0,
+          0, 0, 0, 1,
+        ],
```
kylebarron: Aren't these the default values? We don't need to set these, right? In effect the only thing we care about now is `cartesian` vs `latlng`?

packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
```diff
@@ -187,6 +187,50 @@ export class RasterTileLayer<
...
+    inverseReproject: (x: number, y: number) => [number, number];
+  };
+
+  private _getCommonSpaceReproject(descriptor: RasterTilesetDescriptor): {
+    forwardReproject: (x: number, y: number) => [number, number];
+    inverseReproject: (x: number, y: number) => [number, number];
+  } {
+    if (this._commonSpaceReproject?.descriptor !== descriptor) {
+      const scale = WEB_MERCATOR_TO_WORLD_SCALE;
+      const offset = TILE_SIZE / 2;
```
kylebarron: Why is this divided by 2?

packages/deck.gl-raster/src/raster-layer.ts
```diff
@@ -321,17 +341,32 @@ export class RasterLayer extends CompositeLayer<RasterLayerProps> {
...
+    // dev-docs/specs/2026-05-19-high-zoom-precision-design.md.
+    const xHi = Math.fround(x);
+    const yHi = Math.fround(y);
+    positions[i * 3] = xHi;
+    positions[i * 3 + 1] = yHi;
     // z (flat on the ground)
     positions[i * 3 + 2] = 0;
+    positions64Low[i * 3] = x - xHi;
+    positions64Low[i * 3 + 1] = y - yHi;
+    positions64Low[i * 3 + 2] = 0;
```
kylebarron: I wonder if this would be cleaner if it was refactored into a function named `splitFloat64` that returned `[low, high]`? But in that case will it have a perf impact from array creation?

packages/deck.gl-raster/src/raster-layer.ts
```diff
@@ -293,7 +305,15 @@ export class RasterLayer extends CompositeLayer<RasterLayerProps> {
...
-        data: [1],
+        //
+        // Use the binary { length, attributes } form so we can pipe the
+        // per-vertex fp64 low-part attribute through `data.attributes`.
+        // deck.gl 9.x removed `props.<attributeName>` as a channel for
+        // attribute values, requiring `data.attributes.<attributeName>`
+        // (see @deck.gl/core/lib/attribute/attribute-manager.ts:196).
+        data: positions64Low
+          ? { length: 1, attributes: { positions64Low } }
+          : [1],
```
kylebarron: is positions64Low ever nullish? I thought we were always using 64bit math now?

packages/deck.gl-raster/src/mesh-layer/mesh-layer.ts
```diff
@@ -69,6 +94,13 @@ export class MeshTextureLayer extends SimpleMeshLayer<
     }
 
     super.updateState(params);
+
+    // Dev-mode assertion: the fp64 mesh-vertex correction is only valid for a
+    // single non-instanced mesh with identity per-instance transforms. See
+    // dev-docs/specs/2026-05-19-high-zoom-precision-design.md § Invariant.
+    if (process.env.NODE_ENV !== "production") {
+      assertFp64Invariants(params.props);
+    }
```
kylebarron: This sounds like a code smell to check for NODE_ENV; is this reliable? If this is reliable, then we should make a top-level CONST global variable defined as `const IS_PRODUCTION = true;`

Or we could make this fire only when `debug: true`?

packages/deck.gl-raster/src/mesh-layer/assert-fp64-invariants.ts
```diff
@@ -0,0 +1,83 @@
...
+        "MeshTextureLayer fp64 invariant: getTransformMatrix must not be a function accessor — instance model matrix must remain identity (default: empty array).",
+      );
+    }
+    if (Array.isArray(m) && m.length !== 0) {
+      throw new Error(
+        `MeshTextureLayer fp64 invariant: getTransformMatrix must be left at its default empty value (got length ${m.length}). Any non-empty transform matrix breaks the fp64 correction.`,
+      );
+    }
+  }
+}
```
kylebarron: I think these might be better to just document rather than check at runtime? In particular, we're subclassing the SimpleMeshLayer but we really have a totally different use case than the generic SimpleMeshLayer. We're not rendering 3D models. I think we can make clear that our MeshTextureLayer is spiritually based on the upstream SimpleMeshLayer, but that we explicitly exclude many of the props that the above SimpleMeshLayer supports. We can just remove all of these above props from input.

packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
```diff
@@ -474,17 +478,7 @@ export class RasterTileLayer<
...
-        coordinateOrigin: [0, 0, 0],
-        // biome-ignore format: array
-        modelMatrix: [
-          1, 0, 0, 0,
-          0, 1, 0, 0,
-          0, 0, 1, 0,
-          0, 0, 0, 1,
-        ],
-      };
+      deckProjectionProps = { coordinateSystem: "cartesian" };
```
kylebarron: Instead of making this a poorly typed `Partial<LayerProps>` object; we should just set `coordinateSystem` as a variable

Issue #560, opened 2026-05-20
esm.sh-bundled @developmentseed/deck.gl-geotiff loses proj4 projection classes (utm, lcc, …); breaks CDN / no-build-step consumers


tylere:
## Summary

I'm having problems using @developmentseed/deck.gl-geotiff from esm.sh
\- Tyler

***(The following was written by Claude Code and reviewed by Tyler)***

Loading `@developmentseed/deck.gl-geotiff@0.7.0` from a CDN that aggressively tree-shakes (esm.sh's `?bundle-deps`) drops proj4's non-fundamental projection classes from the bundle, so any COG whose CRS isn't `merc` / `longlat` / `tmerc` fails at the first proj4(sourceProjection, "EPSG:4326") call with:

`Error: Could not get projection name from: [object Object]`

## Repro

```
<script type="module">
  import { COGLayer } from "https://esm.sh/@developmentseed/deck.gl-geotiff@0.7.0?bundle-deps";

  // Any WGS84 UTM COG — e.g. the same Sentinel-2 TCI used in cog-basic.
  const url = "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/18/T/WL/2026/1/S2B_18TWL_20260101_0_L2A/TCI.tif";

  // No `epsgResolver` supplied — uses the package default (epsg.io PROJJSON
  // + wkt-parser), same as examples/cog-basic.
  new COGLayer({ id: "cog", geotiff: url });
</script>
```

Result: the layer initializes, fetches metadata, calls `epsgResolver(32618)`, gets back a definition with `projName: "utm"`, hands it to p`roj4(...)`. proj4 throws — its `Projection.projections` registry doesn't have an entry for "utm".

## Why local builds work and CDN doesn't

The [examples/cog-basic](https://github.com/developmentseed/deck.gl-raster/tree/main/examples/cog-basic) app handles UTM, NZTM, EPSG:2056, etc. with no custom `epsgResolver` and no extra setup. Vite + Rollup keep proj4's projection-class side effects, so the locally-bundled proj4 has a complete `Projection.projections` registry.

esm.sh's `?bundle-deps` runs a more aggressive tree-shaker: from its POV, only `proj4`'s default-exported function is referenced from `@developmentseed/proj` and `cog-layer.ts`, so it eliminates the projection-class registrations as dead code.

I verified this is the failure mode by:

[...]

PR #564, opened 2026-05-21
fix(deck.gl-geotiff): Re-register proj4 projections to survive CDN tree-shaking

tylere:
proj4/lib/index.js calls includedProjections(proj4) at module init to register utm, lcc, tmerc, and the rest of the built-in projection classes. CDN bundlers such as esm.sh and jsDelivr /+esm drop that side-effect call when re-bundling our package, leaving only merc and longlat registered. The result is that any non-trivial COG (UTM, LCC, state-plane, NZTM2000, ...) throws "No code 'foo' found" at runtime when deck.gl-geotiff is loaded from a CDN, while npm/Vite/webpack consumers are unaffected.

Add a dedicated `proj4-init.ts` module that calls
`includedProjections(proj4)` itself, side-effect-import it from `src/index.ts` so every consumer of the package (COGLayer, MosaicLayer, MultiCOGLayer, geotiff helpers) picks it up, and narrow `"sideEffects": false` to `["./dist/proj4-init.js"]` so bundlers preserve only this file's effect and keep tree-shaking everything else.

### DIFF
--- a/packages/deck.gl-geotiff/package.json
+++ b/packages/deck.gl-geotiff/package.json
@@ -11,7 +11,9 @@
       "default": "./dist/index.js"
     }
   },
-  "sideEffects": false,
+  "sideEffects": [
+    "./dist/proj4-init.js"
+  ],
   "files": [
     "dist"
   ],
--- a/packages/deck.gl-geotiff/src/index.ts
+++ b/packages/deck.gl-geotiff/src/index.ts
@@ -1,3 +1,5 @@
+import "./proj4-init.js";
+
 export type {
   COGLayerProps,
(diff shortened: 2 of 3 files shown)


kylebarron:
Sorry, but this looks like a total hack that I don't want to merge. It should not be the domain of deck.gl-geotiff to hack around specific types of bundlers and how they relate to proj4.

kylebarron:
Closing as a bug in upstream esm.sh with suggestions from Claude in the issue here: https://github.com/developmentseed/deck.gl-raster/issues/560#issuecomment-4512056621

PR #572, opened 2026-05-26
fix(geotiff): revert ConcurrencyLimiter to a chunkd SourceMiddleware

### LINKED ISSUE #565: Revert ConcurrencyLimiter to a chunkd SourceMiddleware once chunkd forwards the abort signal
> Written by Claude on behalf of @kylebarron 

## Summary

`GeoTIFF.fromUrl` gates network reads through the `ConcurrencyLimiter` by wrapping the source (`LimitedSource` in `packages/geotiff/src/limiter.ts`) and composing `SourceChunk` / `SourceCache` on top of it. This is a workaround: the natural design is a chunkd `SourceMiddleware` (composes directly with the other middleware), but chunkd's `SourceView` does not forward the request `signal` to middleware, so a middleware can't observe an abort — only the underlying source receives the read options (incl. `signal`) via `SourceView`'s terminal handler.

Upstream fix: https://github.com/blacha/chunkd/pull/1697

## When that ships and we bump `@chunkd/source`

Revert `LimitedSource` back to a `SourceMiddleware` (`LimiterMiddleware`) and compose it into the header `SourceView` after `SourceChunk` / `SourceCache` (so cache hits still short-circuit). The pre-workaround implementation is in git history (commit before this issue's fix) — `git show` the `LimiterMiddleware` class in `packages/geotiff/src/limiter.ts` and the `fromUrl` composition in `packages/geotiff/src/geotiff.ts`.

This is an internal-only change — the public API (`GeoTIFF.fromUrl` options, `ConcurrencyLimiter`, `Priority`, `PerOriginSemaphore`) does not change either way.

## Bonus (also fixed by #1697)

chunkd's `SourceChunk.fetchChunks` dropped the signal on multi-chunk sub-requests; #1697 forwards it. Until released, multi-chunk header reads gate without a signal (rare — header tags are small/single-chunk).

Context: split out of #557.

kylebarron:
> Written by Claude on behalf of @kylebarron

Closes #565.

## Summary

chunkd [#1697](https://github.com/blacha/chunkd/pull/1697) shipped (`@chunkd/source@11.4.1`, `@chunkd/middleware@11.3.1`), so `SourceView` now forwards the request `signal` to middleware (and `SourceChunk` forwards it on multi-chunk sub-requests). That removes the reason for the `LimitedSource` source-wrapper workaround introduced in #557: a middleware can now observe an abort, so a request whose caller aborts while queued for a slot is dropped before any network I/O.

This reverts the limiter to the originally-intended `LimiterMiddleware` (a chunkd `SourceMiddleware`):

- **`limiter.ts`** — `LimitedSource` → `LimiterMiddleware`.
- **`geotiff.ts`** — compose the limiter into the header `SourceView` *after* `SourceChunk`/`SourceCache` (cache hits short-circuit and never burn a slot), plus a tiny `SourceView` wrapping the data source. `getPriority` threading preserved.
- **deps** — bump `@chunkd/source` / `@chunkd/source-http` to `^11.4.1` and `@chunkd/middleware` to `^11.3.1` (the middleware approach hard-requires the fix, so the floors move, not just the lockfile).
- **spec** — drop the temporary "implementation note" in `dev-docs/specs/2026-05-19-concurrency-limiter-design.md`.

**Internal-only:** the public API (`GeoTIFF.fromUrl` options, `ConcurrencyLimiter`, `Priority`, `PerOriginSemaphore`) is unchanged either way.

## Not a blind revert

Two changes landed on `main` *after* the original workaround, so a wholesale `git revert` would have dropped them. Both are preserved:

- the `normalizePriority` (NaN) helper in `limiter.ts`, and
- the integration test that aborts a queued header read through the real `fromUrl → SourceView → limiter` path.

That integration test is the acceptance test for this change — it is unchanged and still passes, which is the end-to-end proof that the signal now reaches the middleware.

## Verification

- `pnpm --filter @developmentseed/geotiff test` — 140 passed (16 files)
- `pnpm --filter @developmentseed/geotiff typecheck` — clean
- `pnpm check` (biome) — clean

### DIFF
--- a/dev-docs/specs/2026-05-19-concurrency-limiter-design.md
+++ b/dev-docs/specs/2026-05-19-concurrency-limiter-design.md
@@ -112,8 +112,6 @@ class LimiterMiddleware implements SourceMiddleware {
 
 `LimiterMiddleware` is also internal: the limiter is wired by `GeoTIFF.fromUrl` (where chunkd's `Source` / `SourceView` types are already in scope), so callers don't need to compose middleware themselves.
 
-> **Implementation note (temporary).** The `SourceMiddleware` shape above is the intended design, but it doesn't work yet: chunkd's `SourceView` doesn't forward the request `signal` to middleware, so a middleware can't observe an abort (only the underlying source receives the read options via `SourceView`'s terminal handler). Until that's fixed upstream ([chunkd#1697](https://github.com/blacha/chunkd/pull/1697)), the limiter is implemented as an internal **source wrapper** — `LimitedSource` — composed *beneath* `SourceChunk` / `SourceCache` (as the `SourceView`'s source) rather than as a middleware on top. Cache hits still short-circuit in `SourceCache` before reaching it. This is internal-only; the public API is unchanged either way. Revert to `LimiterMiddleware` once chunkd ships the fix — tracked in [#565](https://github.com/developmentseed/deck.gl-raster/issues/565).
-
 ## Integration
 
 ### `@developmentseed/geotiff`
--- a/packages/geotiff/package.json
+++ b/packages/geotiff/package.json
@@ -50,9 +50,9 @@
     "extends": "../../package.json"
   },
   "dependencies": {
-    "@chunkd/middleware": "^11.3.0",
-    "@chunkd/source": "^11.4.0",
-    "@chunkd/source-http": "^11.4.0",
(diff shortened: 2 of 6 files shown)


kylebarron (commented):
LGTM!

PR #576, opened 2026-05-27
docs(specs): add antimeridian seam-handling detail

kylebarron:
Splitting at \xc2\xb1180\xc2\xb0 isn't quite enough: proj4 normalizes the seam to the +max_X
boundary, so the wrapped (negative-side) piece's seam corner lands at common-x
512 while its interior is near 0 -> still spans the world -> diverges. Fix:
on the wrapped piece's forwardReproject, if the projected X comes back positive,
subtract one world-width (output-sign-based, deterministic, only the seam flips).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/deck.gl-raster/src/raster-tileset/antimeridian-cut.ts
+++ b/packages/deck.gl-raster/src/raster-tileset/antimeridian-cut.ts
@@ -12,15 +12,44 @@ export interface AntimeridianCut {
   uCut: number;
 }
 
+/**
+ * Tolerance for treating the top- and bottom-edge crossings as the same u
+ * (i.e. the cut is vertical). Crossings further apart than this are rejected as
+ * slanted.
+ */
 const U_EPSILON = 1e-6;
+
+/**
+ * Locate where a single horizontal edge crosses the antimeridian, as a fraction
+ * of the edge's eastward span (0 at the west corner, 1 at the east corner).
+ *
+ * Returns `undefined` if the edge does not cross: with u increasing eastward, a
+ * non-crossing edge has west lng < east lng, while a crossing edge wraps, so
+ * `eastLng < westLng`.
+ */
+function edgeUCut(westLng: number, eastLng: number): number | undefined {
+  // Not crossing if the eastward span doesn't wrap.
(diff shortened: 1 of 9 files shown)


packages/deck.gl-raster/src/raster-tileset/antimeridian-cut.ts
```diff
@@ -0,0 +1,59 @@
...
+  bottomRight: number;
+}
+
+/** A vertical antimeridian cut in a tile's UV space. */
+export interface AntimeridianCut {
+  /** UV u-coordinate (0..1) where the tile crosses ±180°. */
+  uCut: number;
+}
+
+const U_EPSILON = 1e-6;
```
kylebarron, line 27: We should add a small docstring for this

packages/deck.gl-raster/src/raster-tileset/antimeridian-cut.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    // Eastward distance west→(+180) then (−180)→east.
+    const toSeam = 180 - westLng;
+    const fromSeam = eastLng + 180;
+    const total = toSeam + fromSeam;
+    if (total <= 0) {
+      return undefined;
+    }
+    return toSeam / total;
+  };
```
kylebarron: I think this can be a top-level function

packages/deck.gl-raster/src/raster-tileset/antimeridian-cut.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const topUCut = edgeUCut(topLeft, topRight);
+  const bottomUCut = edgeUCut(bottomLeft, bottomRight);
+  if (topUCut === undefined || bottomUCut === undefined) {
+    return undefined;
+  }
+  // Vertical only: both edges must cross at the same u.
```
kylebarron: This comment should be expanded to say that we _should_ support more general antimeridian handling, we just don't right now

PR #579, opened 2026-05-28
fix: Support parsing ESRI WKT String for ESRI-created custom CRS

kylebarron:
Ref https://github.com/source-cooperative/cog-viewer/issues/8 and https://github.com/developmentseed/geotiff-test-data/pull/58 and https://github.com/developmentseed/async-geotiff/pull/165. This GeoTIFF includes

GeoTIFF has `model_type`, `geographic_type`, `projected_type` geokeys. 

- If `model_type` is 1, check the `projected_type`
    - If `projected_type` is an int < 32767, it's an EPSG code; else CRS encoded in GeoKeys
- If `model_type` is 2, check the `geographic_type`
    - If `geographic_type` is an int < 32767, it's an EPSG code; else CRS encoded in GeoKeys
- If `model_type` is 32767, then the projection is "undefined", and there's a (hack? grandfathered?) way that uses ESRI WKT in the proj citation key.

This PR now supports the latter approach when `model_type` itself is 32767.

### DIFF
--- a/packages/geotiff/src/array.ts
+++ b/packages/geotiff/src/array.ts
@@ -43,8 +43,13 @@ export type RasterArrayBase = {
    */
   transform: Affine;
 
-  /** Coordinate reference system information. */
-  crs: number | ProjJson;
+  /** Coordinate reference system information.
+   *
+   * - If `crs` is a number, it is an EPSG code.
+   * - If `crs` is an object, it is a PROJJSON object.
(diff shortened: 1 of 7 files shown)


kylebarron:
The image still doesn't _render_ but this is a valid improvement (with a test), so we can move the rendering question to the next branch.

PR #602, opened 2026-06-12
docs: Add git submodule install instructions

abarciauskas-bgse:
While attempting to run the examples locally, I found a step was missing (I think?) from the setup instructions which is to install git submodules. This PR adds instructions for doing install git submodules.

I also thought it might be easier in the future to maintain all setup instructions in one place and link to that on the individual examples pages, but I can revert that change if you like have standalone READMEs for each example. (Or if you like that change I can apply it across examples)

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -2,6 +2,14 @@
 
 This is a monorepo managed with pnpm workspaces.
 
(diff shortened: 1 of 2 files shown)


kylebarron (approved):
thanks!

PR #613, opened 2026-07-03
fix: Plumb material prop to MeshTextureLayer so rasters can render unlit

### LINKED ISSUE #612: Raster imagery darkened by scene LightingEffect; no consumer control over mesh material/lighting
## Problem

Image

Raster imagery rendered through `RasterLayer`/`MeshTextureLayer` (and therefore `COGLayer`) is uniformly darkened whenever the deck.gl scene has a `LightingEffect` with ambient intensity < 1 — a common setup when the same scene lights 3D extrusions. Rendered pixels come out at roughly `fileValue × ambientIntensity` (we measured 40% and 60% in two views), while `BitmapLayer`-based imagery in the same scene shows true values. Downloading the GeoTIFF shows correct pixels; only the on-screen render is dark.

## Root cause

The vendored mesh fragment shader (`packages/deck.gl-raster/src/mesh-layer/mesh-layer-fragment.glsl.ts`) runs the sampled color through phong lighting:

```glsl
vec3 lightColor = lighting_getLightColor(color.rgb, cameraPosition, position_commonspace.xyz, normal);
fragColor = vec4(lightColor, color.a * layer.opacity);
```

`MeshTextureLayer`'s default `material: { ambient: 1.0, diffuse: 0.0, ... }` intends to neutralize lighting, but the phong ambient term is `material.ambient × ambientLight.color × ambientLight.intensity × surfaceColor` — it's only neutral when the scene's ambient intensity is exactly 1.0.

There is also no consumer-facing escape hatch: the mesh layer sits behind `COGLayer → TileLayer → RasterLayer → MeshTextureLayer`, the `TileLayer` is constructed bare, and per-tile sublayer ids are dynamic, so `_subLayerProps` can't reach it with a `material` override.

## Related deck.gl v9 regression (affects the fix design)

Even where `material` *can* be passed (directly to `MeshTextureLayer`), deck.gl's documented `material: false` ("disable lighting") is silently broken in v9:

- `LightingEffect.getShaderModuleProps()` forwards the layer's raw `material` prop as the `phongMaterial` module props (`@deck.gl/core` `effects/lighting/lighting-effect.ts`).
- luma.gl's `ShaderInputs.setProps()` coerces falsy module props to `{}`:
  ```js
  const moduleProps = props[moduleName] || {};
  ```
  so `false` → `{}` → `getUniforms({})` → the **default lit** phong material (`ambient 0.35, diffuse 0.6`) — arguably darker still.

The unlit escape hatch exists in the shader (`if (material.unlit) return surfaceColor;` in `@luma.gl/shadertools` phong module); it's just unreachable via `material: false` in v9.

## Proposal

Keep the current default (no visual change for existing consumers), and expose deck.gl's standard `material` prop on `RasterTileLayer` and `RasterLayer`, forwarded conditionally down to `MeshTextureLayer` (following the existing `debugOpacity` pattern; forwarded only when set so the current default is preserved). `COGLayer` extends `RasterTileLayer`, so it inherits the prop with zero deck.gl-geotiff changes.

To make `material: false` actually mean "unlit" despite the v9 regression above, `MeshTextureLayer.draw()` (which already sets per-frame shader module props after the effect's props are applied) translates a falsy material into `phongMaterial: { unlit: true }`.

Result: `new COGLayer({ ..., material: false })` renders file values verbatim — parity with `BitmapLayer`.

We considered a raster-specific boolean (e.g. `lighting: false`) instead; `material` follows the `SimpleMeshLayer` convention and gives custom-material control for free, but happy to rename if you prefer a raster-specific term.

**PR incoming from us implementing the above.**

## Workaround (0.7.x)

pnpm patch on `dist/mesh-layer/mesh-layer-fragment.glsl.js` replacing the lit `fragColor` with `vec4(color.rgb, color.a * layer.opacity)`.

charlieforward9:
Closes #612

## Problem

Raster imagery drawn through `MeshTextureLayer` (so `RasterLayer`, `RasterTileLayer`, and `COGLayer`) is uniformly darkened whenever the deck.gl scene has a `LightingEffect` with ambient intensity < 1 — a common setup when the same scene lights 3D extrusions. Rendered pixels come out at roughly `fileValue × ambientIntensity` (we measured 40% and 60% in two production views), while `BitmapLayer` imagery in the same scene shows true values.

The default `material: { ambient: 1.0, diffuse: 0.0, ... }` intends to neutralize lighting, but the phong ambient term is `material.ambient × ambientLight.color × ambientLight.intensity × surfaceColor` — neutral only when the scene's ambient intensity is exactly 1.0. And there is no consumer-facing escape hatch: `_subLayerProps` can't reach the mesh layer through the tile-layer indirection.

## Change

- **`RasterTileLayer` / `RasterLayer`**: new optional `material?: Material` prop (deck.gl's standard type; `false` = unlit), forwarded down the chain **only when set** (`...(material !== undefined && { material })`) so the current default behavior is untouched for existing consumers. `COGLayer` inherits the prop via `RasterTileLayer` with no deck.gl-geotiff changes.
- **`MeshTextureLayer.draw()`**: translates a falsy `material` into `phongMaterial: { unlit: true }` shader-module props.

The translation is needed because deck.gl v9 regressed `material: false`: `LightingEffect.getShaderModuleProps()` forwards the layer's raw `material` prop as the `phongMaterial` module props, and luma.gl's `ShaderInputs.setProps()` coerces falsy module props to `{}` —

```js
const moduleProps = props[moduleName] || {};
```

— which re-applies the **default lit** phong material (`ambient 0.35, diffuse 0.6`). The unlit escape hatch already exists in the shader (`if (material.unlit) return surfaceColor;` in `@luma.gl/shadertools`); it's just unreachable via `material: false` in v9. `MeshTextureLayer.draw()` already sets per-frame module props after the effect's props are applied (`Layer._drawLayer` order), so the translation deterministically wins. This also fixes `material: false` for anyone passing it directly to `MeshTextureLayer` today.

Result:

```ts
new COGLayer({ ..., material: false })  // renders pixel values verbatim, like BitmapLayer
```

We considered a raster-specific boolean (e.g. `lighting: false`) instead — happy to rename, but `material` follows the `SimpleMeshLayer` convention and gives custom-material control for free.

## Testing

- `tests/raster-layer.test.ts`: forwards `material: false` / material object to `MeshTextureLayer`; key **absent** (not `undefined`) when unset, pinning the default-preserving conditional spread.
- `tests/raster-tile-layer/material-forwarding.test.ts` (new): same assertions across `RasterTileLayer._renderSubLayers` → `RasterLayer`.
- `tests/mesh-layer.test.ts` (new): `material: false` → `setProps` called with `phongMaterial: {unlit: true}`; default and custom materials leave `phongMaterial` untouched.

`pnpm --filter @developmentseed/deck.gl-raster test` (146 passed), `pnpm lint`, `pnpm --filter @developmentseed/deck.gl-raster typecheck` all green (the `docs` workspace typecheck failure pre-exists on main — missing generated typedoc sidebar).

Verified end-to-end in our production app (AGROZOOM COG imagery over a scene lit at ambient 0.4/0.6): `material: false` renders identically to a patched unlit shader; omitting the prop reproduces today's behavior exactly.

### DIFF
--- a/packages/deck.gl-raster/src/mesh-layer/mesh-layer.ts
+++ b/packages/deck.gl-raster/src/mesh-layer/mesh-layer.ts
@@ -175,6 +175,19 @@ export class MeshTextureLayer extends SimpleMeshLayer<
       shaderProps[m.module.name] = m.props || {};
     }
 
+    // deck.gl documents `material: false` as the way to render a mesh unlit,
+    // but in deck.gl v9 the LightingEffect forwards the raw `false` to
+    // luma.gl's ShaderInputs, which coerces falsy module props to `{}` and so
+    // silently re-applies the *default* phong material (ambient 0.35,
+    // diffuse 0.6). Translate a disabled material into the phongMaterial
+    // module's `unlit` flag ourselves so raster values render verbatim (like
+    // `BitmapLayer`). This runs after the LightingEffect's module props are
+    // applied for the frame (`Layer._drawLayer` sets effect shaderModuleProps
+    // before calling `draw`), so it reliably wins.
+    if (!this.props.material) {
+      shaderProps.phongMaterial = { unlit: true };
+    }
+
     for (const m of super.getModels()) {
       m.shaderInputs.setProps(shaderProps);
     }
--- a/packages/deck.gl-raster/src/raster-layer.ts
+++ b/packages/deck.gl-raster/src/raster-layer.ts
@@ -2,6 +2,7 @@ import type {
   CompositeLayerProps,
   DefaultProps,
   Layer,
+  Material,
   TextureSource,
   UpdateParameters,
 } from "@deck.gl/core";
@@ -119,6 +120,21 @@ export interface RasterLayerProps extends CompositeLayerProps {
 
   /** Opacity of the debug overlay. */
   debugOpacity?: number;
+
+  /**
+   * Lighting material forwarded to the underlying {@link MeshTextureLayer}
+   * (a `SimpleMeshLayer` subclass). Follows deck.gl's `material` convention:
+   * pass `false` to render the raster unlit, so pixel values display verbatim
+   * even when the scene has a `LightingEffect` with ambient intensity < 1
+   * (matching `BitmapLayer` behavior).
+   *
+   * When omitted, `MeshTextureLayer`'s default material applies
+   * (`{ambient: 1, diffuse: 0, shininess: 0, specularColor: [0, 0, 0]}`),
+   * which is lighting-neutral only when the scene's ambient intensity is 1.
+   *
+   * @default undefined
+   */
+  material?: Material;
 }
 
 const defaultProps: DefaultProps<RasterLayerProps> = {
@@ -333,7 +349,7 @@ export class RasterLayer extends CompositeLayer<RasterLayerProps> {
 
   renderLayers() {
     const { mesh, positions64Low } = this.state;
-    const { debug, image, renderPipeline } = this.props;
+    const { debug, image, renderPipeline, material } = this.props;
 
     // mesh and positions64Low are always set together by _generateMesh.
     if (
@@ -349,6 +365,10 @@ export class RasterLayer extends CompositeLayer<RasterLayerProps> {
         id: "raster",
         image,
         renderPipeline,
+        // Forward only when set: an explicit `material: undefined` own-prop
+        // would shadow MeshTextureLayer's default material (deck.gl props are
+        // prototype-chained onto defaultProps).
+        ...(material !== undefined && { material }),
         // Single mesh rendered as one non-instanced draw.
         data: { length: 1, attributes: { positions64Low } },
         mesh,
--- a/packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
+++ b/packages/deck.gl-raster/src/raster-tile-layer/raster-tile-layer.ts
@@ -3,6 +3,7 @@ import type {
   CoordinateSystem,
   DefaultProps,
   Layer,
+  Material,
 } from "@deck.gl/core";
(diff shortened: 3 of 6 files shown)


packages/deck.gl-raster/src/mesh-layer/mesh-layer.ts
```diff
@@ -175,6 +175,19 @@ export class MeshTextureLayer extends SimpleMeshLayer<
...
+    // silently re-applies the *default* phong material (ambient 0.35,
+    // diffuse 0.6). Translate a disabled material into the phongMaterial
+    // module's `unlit` flag ourselves so raster values render verbatim (like
+    // `BitmapLayer`). This runs after the LightingEffect's module props are
+    // applied for the frame (`Layer._drawLayer` sets effect shaderModuleProps
+    // before calling `draw`), so it reliably wins.
+    if (!this.props.material) {
+      shaderProps.phongMaterial = { unlit: true };
+    }
+
```
kylebarron, line 190: this seems like an upstream bug? Do you expect third party layers to have to know all this?

kylebarron:
Unless you can explain in plain language, without AI, what the problem is, we need this change, how we can test it, and how we can avoid regressions in the future, I'm going to close this. Because I don't understand what it's doing.

charlieforward9:
> The COG overlay rendered 40-60% darker than its file values: deck.gl-raster's MeshTextureLayer runs imagery through deck.gl phong lighting, and our LightingEffect (ambient 0.4-0.6 for 3D tree columns) multiplies straight into the raster. PNG ortho tiles use BitmapLayer (unlit) — hence the layer-to-layer mismatch.

Not saying this is the fix, but this is what the agent was able to determine as I was adopting this package in favor of handling my own TIF tiling & rendering.

Issue #614, opened 2026-07-05
`ZarrLayer` parses but ignores `spatial:registration="node"`


abhi-airspace-intelligence:
## Background

I'm trying to visualize the NOAA GFS layer, which I've converted into a GeoZarr-aligned / Icechunk dataset. Since GFS publishes weather nodes, I tried to pass `spatial:registration="node"` to the `ZarrLayer`, but unfortunately, it appears that ZarrLayer ignores this entirely, which causes it to render off-center.

abhi-airspace-intelligence:
Planning on fixing this bug today

kylebarron:
I think you just need this line:
https://github.com/developmentseed/deck.gl-raster/blob/c1230c9d0ddf4bc67b340e8554ef778778025306/packages/geotiff/src/transform.ts#L28-L31
`RasterTypeKey.PixelIsPoint` is the GeoTIFF version of `spatial:registration="node"` I think

abhi-airspace-intelligence:
@kylebarron thanks! Do you know when they'll be a new release? Would love to pick up this fix and #590?

PR #615, opened 2026-07-06
fix: handle `spatial:registration = node` in geozarr properties

abhi-airspace-intelligence:
Addresses #614 

This takes the logic from the GeoTiff pixel area handling, extracts it to the affine shared package, and then applies it to the Zarr handling when `spatial:registration = node` is set. Previously, the property was parsed, but not used for anything.


kylebarron (approved):
Looks good; thanks!

Issue #159, opened 2026-01-06
Handle band-interleaved/band-separate data (e.g. landsat/sentinel)


kylebarron:
chatgpt nudged me towards texture2darray here, which is actually brilliant, especially this:

> Data-driven band selection (texture array model)
> 
> ```glsl
> uniform sampler2DArray bands;
> uniform int bandR;
> uniform int bandG;
> uniform int bandB;
> 
> float r = texture(bands, vec3(uv, bandR)).r;
> float g = texture(bands, vec3(uv, bandG)).r;
> float b = texture(bands, vec3(uv, bandB)).r;
> ```

Now:
	•	shader stays fixed
	•	band indices are uniforms
	•	change behavior by updating numbers

No recompilation. No new programs.

kylebarron:
Duplicate of https://github.com/developmentseed/deck.gl-raster/issues/142 which was closed by #408, #410, #411, #413, #414

PR #635, opened 2026-08-12
fix(deck.gl-geotiff): render band-separate (PlanarConfiguration=2) images

thatbudakguy:
## What

`getTileData` in `createUnormPipeline` throws for any COG stored with `PlanarConfiguration=2` (GDAL's `INTERLEAVE=BAND`):

```js
if (array.layout === "band-separate") {
  throw new Error("Band-separate images not yet implemented.");
}
```

Because it throws per tile inside the tile loader, a `COGLayer` over such a file draws nothing at all and the only sign is `onTileError` firing once per tile.

This interleaves the tile instead. `toPixelInterleaved` in `@developmentseed/geotiff` already does exactly this — the array helpers just weren't exported from that package's index, although their types were.

## Why this came up

A librarian at UMN reported a scanned map that wouldn't display in [a viewer I'm working on that uses this library](https://github.com/OpenGeoMetadata/ogm-viewer/issues/158). The file is a 4-band RGBA JPEG COG with an alpha collar, written with `-of COG -co COMPRESS=JPEG -co INTERLEAVE=BAND`, and the reader handles it fine — only the render pipeline refuses it.

## Changes

- **`packages/geotiff`** — export `toPixelInterleaved`, `toBandSeparate`, `reorderBands`, `packBandsToRGBA` and `PackBandsToRGBAOptions` from the index. The `RasterArray*` types were already exported; the functions that produce them weren't.
- **`render-pipeline.ts`** — interleave a band-separate tile before the RGB padding step, rather than throwing. Doing it first means the padding and texture-format logic below see the same shape they would for a pixel-interleaved image, so 3-band `INTERLEAVE=BAND` is fixed alongside 4-band.
- **`geotiff.ts`** — `addAlphaChannel` carried a second copy of the same throw, and unlike the pipeline it's part of the public API. It now interleaves too. Its return type narrows from `RasterArray` to `RasterArrayPixelInterleaved`, which every path through it has always satisfied; that's what lets the narrowing in `getTileData` survive the call.
- **`examples/vermont-cog-comparison`** — the narrowed return type makes that example's `layout === "band-separate"` guard statically unreachable, so it's removed.

## Tests

- `tests/render-pipeline.test.ts` — `getTileData` interleaves a 4-band tile into one `rgba8unorm` texture, and interleaves *and* pads a 3-band one. The tile is handed in directly rather than read from a fixture: the only band-separate fixture is `int8_3band_zstd_block64`, and `inferRenderPipeline` doesn't build a pipeline for non-unsigned samples. Happy to add a `uint8` band-separate fixture to `geotiff-test-data` as a follow-up if you'd like the coverage to be end-to-end.
- `tests/geotiff.test.ts` (new) — `addAlphaChannel` over band-separate RGB and RGBA input, plus the existing pixel-interleaved path.

I also checked the interleave against real data outside the suite: for a band-separate DEFLATE COG built from the reported file, `toPixelInterleaved` output is byte-identical to `geotiff.js`'s independent `readRasters({ interleave: true })` across four full 256×256 tiles (~1M samples), with the alpha channel spanning 0–255 rather than a constant. And with this change the reported COG renders correctly in ogm-viewer, transparent collar included.

`pnpm build`, `pnpm -r --filter='!docs' typecheck`, `biome ci .` and the `geotiff` / `deck.gl-geotiff` / `deck.gl-raster` suites all pass locally. The 17 failures in `packages/geotiff/tests/integration-rasterio.test.ts` are the `.npy` reference fixtures that CI generates with `pixi run generate-npy`; I don't have pixi here, and they fail the same way on a clean checkout.

## Note on approach

Uploading one texture per band and sampling a `sampler2DArray` — the idea in #159 — would avoid the per-tile copy this adds, and is the better long-term shape. It needs the render pipeline to know how many textures a tile has, which is a larger change. This is the fallback path `packBandsToRGBA`'s docstring already describes ("useful as a fallback path when a single 4-channel texture upload is preferred over one texture per band"), and it makes these files render today. Glad to rework it if you'd rather go straight to the texture-array version.


thatbudakguy:
Apologies for the largely AI-generated PR; this came up in some work I was doing and seemed like a straightforward fix. 

Let me know if these changes aren't desirable!

Issue #625, opened 2026-07-23
Adaptive reprojection mesh never converges near the poles for global EPSG:4326 sources → high-latitude shear


tylere:
## Summary

When a global **EPSG:4326** raster is rendered on a Web-Mercator map, tiles at high latitudes are visibly **sheared / streaked**; equatorial and mid-latitude tiles are correct. The cause is `@developmentseed/raster-reproject`'s `RasterReprojector` (used by `RasterLayer._generateMesh`): its adaptive mesh **fails to converge near the poles**, hits the `maxIterations` cap (10000), and breaks — leaving a badly under-resolved mesh for that tile.

This is already flagged in the code itself — `RasterReprojector.run()`:

```js
// Note: this primarily happens near the poles, where we'll essentially
// never converge
while (this.getMaxError() > maxError) {
  this.refine();
  if (++iterations > maxIterations) {
    console.warn(`RasterReprojector: mesh refinement did not converge ...`);
    break;
  }
}
```

## Minimal repro (headless, uses this repo's own classes)

```js
// deps: @developmentseed/deck.gl-raster, @developmentseed/raster-reproject,
//       @developmentseed/proj, proj4
import proj4 from "proj4";
import { AffineTilesetLevel } from "@developmentseed/deck.gl-raster";
import { RasterReprojector } from "@developmentseed/raster-reproject";
import { makeClampedForwardTo3857 } from "@developmentseed/proj";

// A plain global EPSG:4326 grid: 0.1°/px, 3600×1800, 256² tiles (7 tile rows).
const level = new AffineTilesetLevel({
  affine: [0.1, 0, -180, 0, -0.1, 90], // [scaleX,0,originX,0,scaleY,originY]
  arrayWidth: 3600, arrayHeight: 1800, tileWidth: 256, tileHeight: 256,
  mpu: 111319.49,
});
const conv = proj4("EPSG:4326", "EPSG:3857");
const to3857 = makeClampedForwardTo3857((x, y) => conv.forward([x, y]), (x, y) => [x, y]);
const from3857 = (x, y) => conv.inverse([x, y]);

function stats(row, label) {
  const { forwardTransform, inverseTransform } = level.tileTransform(0, row);
  let warned = false; const w = console.warn; console.warn = () => (warned = true);
  const rp = new RasterReprojector(
    { forwardTransform, inverseTransform, forwardReproject: to3857, inverseReproject: from3857 },
    257, 257,
  );
[...]

PR #636, opened 2026-08-20
fix: render COGs in bounded projections (Mollweide, Sinusoidal, Equal Earth)

tylere:
## Summary

Fixes rendering of COGs that use a CRS with a bounded domain — an ellipse or other closed shape — such as World Mollweide (ESRI:54009), Sinusoidal, and Equal Earth. Before this PR, loading any such COG produced a blank map.

**Abbreviations used below:**
- **OOD** — Out Of Domain: a point that lies outside the projection's valid region (e.g. a rectangle corner that falls outside the Mollweide ellipse). proj4 returns NaN or a clamped/incorrect value for these points.
- **OBB** — Oriented Bounding Box: a tight-fitting box that can be rotated to align with a shape. Used by deck.gl for frustum culling — testing whether a tile's OBB intersects the camera frustum to decide if the tile is visible.
- **AABB** — Axis-Aligned Bounding Box: a bounding box whose sides are parallel to the coordinate axes (`[minX, minY, maxX, maxY]`). Used here as the `commonSpaceBounds` for a quick pre-filter before the full frustum cull.

Four root causes were identified and fixed:

### 1. `AffineTilesetLevel.projectedTileCorners` — corner extrapolation past array boundary

The last tile in each row/column has nominal corners at `(col+1)*tileWidth` pixels, which exceeds `arrayWidth` when the tile matrix doesn't evenly divide the array. For bounded projections, those extrapolated corners fall outside the CRS domain; proj4 clamps them to ±85.05° in Mercator, collapsing all such corners onto a single horizontal line and producing a zero-height bounding volume — causing **every tile to be culled immediately**.

**Fix:** clip `right = min((col+1)*tw, arrayWidth)` and `bottom = min((row+1)*th, arrayHeight)` before applying the affine.

### 2. `RasterTileNode._getGenericBoundingVolume` — NaN positions in OBB input

OOD corners produce NaN common-space positions. Passing NaN to `makeOrientedBoundingBoxFromPoints` produced a degenerate OBB that always returned -1 from `computeVisibility`, incorrectly culling every tile. Additionally, when the only valid reference points are collinear (e.g. the equatorial band of a Mollweide tile), the OBB is degenerate for a different reason.

**Fix:** filter NaN positions before OBB/AABB computation; return an off-screen degenerate bounding volume for fully-OOD tiles; build the OBB from the four AABB corners (with a 1-unit minimum extent on each axis) so the OBB is always non-degenerate.

### 3. `RasterLayer._generateMesh` — adaptive reprojector choked by OOD vertices

Tiles that straddle the CRS domain boundary have some corners with valid output positions and some OOD. The adaptive `RasterReprojector` spent its entire iteration budget refining OOD-vertex triangles, leaving the valid area of the tile severely under-resolved (error 100s of pixels).

**Fix:** detect "border tiles" by checking whether any corner's round-trip error via `forwardReproject → inverseReproject` exceeds 1% of the tile CRS width. For border tiles, build a dense 64×64 uniform grid mesh and filter out any triangle whose vertex fails the OOD check — producing a clean domain-boundary cutoff. Also filter OOD triangles from the adaptive mesh path in `reprojectorToMesh`.

### 4. `RasterTileset2D.projectPosition` — explicit NaN for OOD inputs

Made the NaN pass-through explicit so the GPU rasterizer reliably discards triangles touching OOD vertices, regardless of future changes to the clamping path.

## Test plan

- [x] `pnpm -r test` — 135 pass (4 pre-existing failures in `web-mercator-clamp.test.ts` related to `triangulateRectangle` not exported from dist; unrelated to this PR)
- [x] `pnpm biome check` — clean across all changed files
- [x] `pnpm -r typecheck` — same pre-existing failures as on `main` (`InitialTriangulation` not in dist); no new errors from this PR
- [x] +2 new edge-clipping tests in `affine-tileset-level.test.ts`
- [ ] Manual: load `https://data.source.coop/vizzuality/hfp-100/hfp_2017_100m_v1-2_cog.tif` (World Mollweide HFP COG) in the `cog-basic` example — tiles should render over the globe

## Related

- Closes #580 (supersedes the `affine-tileset-level` fix from that PR; this PR includes that fix plus the three additional root causes)
- Related to #625 (adaptive reprojector convergence near domain boundaries)


tylere:
These changes are based on patches used with deck.gl-raster 0.8.0-beta.2
See: https://github.com/source-cooperative/cog-viewer/pull/30

kylebarron:
I don't know enough about these types of projections but we definitely need a careful understanding of how they work and careful tests before merging

PR #639, opened 2026-08-21
fix(mosaic): Make MosaicLayer repeat over world copies (antimeridian)

willjnz:
Hey @kylebarron, great work - thank you very much.

I tried the beta v0.8 because I needed what I thought was fixed by #518. It turns out that fix doesn't make `MosaicLayer` repeat across world copies. This PR applies the same fix to `MosaicLayer`.

There are two independent layers of "which things are visible" in this codebase:

1. **Inside an already-open COG**: `RasterTileLayer` decides which pixel tiles `(x, y, z)` of that one file are visible. Fixed by #518.
2. **One level up**: `MosaicLayer`'s `MosaicTileset2D` decides which sources (whole COGs - e.g. STAC items) are visible at all, via a spatial query over their bounding boxes, *before* any `RasterTileLayer` for that source exists.

#518 only fixed layer 1. If a source is never selected at layer 2, it never gets a `RasterTileLayer` built for it (`renderSource` never runs), so #518's fix never gets a chance to run.

The root cause: `MosaicTileset2D.getTileIndices` queried its Flatbush spatial index once, using the viewport's raw (possibly unwrapped) longitude bounds. Once a `WebMercatorViewport` with `repeat: true` pans past ±180°, `viewportBounds` extends past that range while sources are still indexed at their true, unwrapped bbox - so items on the far side of the antimeridian were never selected.

The fix repeats the query at world-copy offsets - same `subViewports.length > 1` gate and `MAX_MAPS` walk-until-empty pattern as `raster-tile-traversal.ts` - shifting the *search bounds* by `worldOffset * 360°` of longitude instead of translating a bounding volume, since Flatbush operates on WGS84 degrees, not deck.gl common space. Documented as a second, independent layer in `dev-docs/world-copies.md`.

## Test example

| Before (v0.8.0-beta.2) | After (this PR) |
|---|---|
| before | after 1<br>after 2 |

**Note:** there is still a gap right at the antimeridian, because COGs with bounds of `-180` to `+180` (antimeridian-spanning) are excluded - a known separate issue, unrelated to this fix. The geomad layer has some COGs that render badly before and after this PR for that reason; LULC does not.

### Minimal app to test

```tsx
import { COGLayer, MosaicLayer } from "@developmentseed/deck.gl-geotiff";
import type { StyleSpecification } from "maplibre-gl";
import "maplibre-gl/dist/maplibre-gl.css";
import { Map as MaplibreMap } from "react-map-gl/maplibre";
import { DeckGLOverlay } from "./deckgl-overlay.js";

type LulcSource = { id: string; bbox: [number, number, number, number]; href: string };

// One COG on each side of the antimeridian (dep_ls_lulc STAC collection).
const SOURCES: LulcSource[] = [
  {
    id: "dep_ls_lulc_063_019_2025",
    bbox: [177.380649859963, -19.29852907893266, 178.24303253271776, -18.4776694259545],
    href: "https://s3.us-west-2.amazonaws.com/dep-public-staging/dep_ls_lulc/0-0-9/063/019/2025/dep_ls_lulc_063_019_2025_classification.tif",
  },
  {
    id: "dep_ls_lulc_067_019_2025",
    bbox: [-179.169819449018, -19.29852907893266, -178.30743677626327, -18.4776694259545],
    href: "https://s3.us-west-2.amazonaws.com/dep-public-staging/dep_ls_lulc/0-0-9/067/019/2025/dep_ls_lulc_067_019_2025_classification.tif",
  },
];

const BASEMAP_STYLE: StyleSpecification = {
  version: 8,
  sources: {
    osm: {
      type: "raster",
      tiles: ["https://tile.openstreetmap.org/{z}/{x}/{y}.png"],
      tileSize: 256,
      attribution: "© OpenStreetMap contributors",
    },
  },
  layers: [{ id: "osm", type: "raster", source: "osm" }],
};

const layer = new MosaicLayer<LulcSource>({
  id: "lulc-mosaic",
  sources: SOURCES,
  renderSource: (source) => new COGLayer({ id: `lulc-${source.id}`, geotiff: source.href }),
});

export default function App() {
  return (
    <div style={{ position: "relative", width: "100%", height: "100%" }}>
      <MaplibreMap
        initialViewState={{ longitude: 180, latitude: -18.9, zoom: 7 }}
        mapStyle={BASEMAP_STYLE}
        renderWorldCopies={true}
      >
        <DeckGLOverlay layers={[layer]} />
      </MaplibreMap>
    </div>
  );
}
```

| Before (v0.8.0-beta.2) | After (this PR) |
|---|---|
| before | after |

### DIFF
--- a/dev-docs/world-copies.md
+++ b/dev-docs/world-copies.md
@@ -79,9 +79,31 @@ Differences:
 - We cache the offset-0 OBB per `(zRange)` and translate from the cache
   rather than recomputing per offset.
 
+## A second, independent layer: MosaicTileset2D source selection
+
+This file focusses primarily on how *pixel tiles* are selected within an already-open COG, inside
+`RasterTileLayer`. A separate concern sits one level up: `MosaicLayer`'s
+[`MosaicTileset2D`](../packages/deck.gl-geotiff/src/mosaic-layer/mosaic-tileset-2d.ts)
+decides which `sources` (each `{ id?, bbox }`, see
+`MosaicSource`) are visible at all via a Flatbush spatial query over source
+bboxes *before* any `RasterTileLayer` for that source exists. `viewport.getBounds()`
+returns unwrapped longitudes once the camera pans past ±180°, so a single,
+un-repeated query misses sources whose bbox sits on the far side of the seam.
+A source that's never selected here never gets a `RasterTileLayer`, so the
+fix above can't help it.
+
+The fix mirrors this doc's model: `subViewports.length > 1` gate and same
+`MAX_MAPS` walk-until-empty pattern but shifts the *query bounds* by
+`worldOffset * 360°` of longitude instead of translating a bounding volume by
+`worldOffset * 512` in common space, since Flatbush operates on WGS84
+degrees, not deck.gl common space. Matches are deduplicated across offsets
+(a `Set`) since the same source can satisfy more than one offset's query at
+once (e.g. a source near the prime meridian at extreme zoom-out).
+
 ## Out of scope
 
 - The large-zarr-root cull path in `createRootTiles` does not yet consider
   world-copy intersections. For typical OGC pyramids this doesn't matter
   (root tiles are enumerated unconditionally). If a zarr dataset hits the
   same symptom, generalize that path.
+- Antimeridian issues such as COGs whose bbox crosses ±180° in WGS84 longitude will be handled separately.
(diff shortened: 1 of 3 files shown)


kylebarron:
Thanks @willjnz; this looks reasonable but I'll need to take a closer look before merging. I'm focused on creating presentations for FOSS4G at the moment

Issue #215, opened 2026-02-03
Runtime Custom Visualization & Texture Filtering Control


MariDani:
Hi everyone,

First of all, I really admire the approach taken in this library with the GPU-based raster reprojection and native tiling - it is a fantastic architecture.

We are currently maintaining our own solution for COG visualization ([gisat/deck.gl-geotiff](https://www.google.com/search?q=https://github.com/Gisat/deck.gl-geotiff)), but we are very interested in potentially migrating to use `deck.gl-raster` in the future. However, we have identified two key features that are currently preventing us from doing so, and I wanted to ask about your roadmap for them.

**1. Runtime Custom Visualization**
Currently, it seems `deck.gl-raster` relies heavily on visualization styles that are saved directly within the GeoTIFF/COG file. For our use cases, we need the ability to define custom visualizations dynamically at runtime (e.g., custom pixel-by-pixel filtering, band combinations, dynamic thresholding, and coloring) without modifying the source file.

We implement this in our current architecture by processing pixel data on the fly (see our [architecture docs here](https://github.com/Gisat/deck.gl-geotiff/blob/dev/geoimage/docs/architecture-geoimage.md)).

I noticed in your README that "spectral band math" and "pixel filtering" are listed under "Soon". Could you share if there is a plan or timeline for enabling these custom visualization pipelines? This would be a major enabler for us to switch.

**2. Texture Filtering Control (Nearest vs. Linear)**
We also need the ability to control texture filtering, specifically to disable texture smoothing (blurring) when zoomed in. For scientific data analysis, preserving the "sharp" pixel edges using Nearest Neighbor interpolation is often critical.

In our current implementation, we handle this by passing `textureParameters` to the layer, toggling between `linear` and `nearest` based on user preference:

````typescript
// Example from our CogBitmapLayer.ts
textureParameters: {
  minFilter: blurredTexture ? 'linear' : 'nearest',
  magFilter: blurredTexture ? 'linear' : 'nearest',
},
````
Is there an option to expose these texture parameters in `COGLayer`?

Thank you for your work on this library!

MariDani:
Hi @kylebarron , thank you very much for the quick response! I successfully tried the texture filtering control. I'll need more time to play with the code to get the custom visualization working, but the path is clear.

Anyway, thank you for your help and good luck with the repository development.

AbelVM:
I'd really love to have something like 
* Pixel data encoded as an array of band values
* `getPixelColor` prop 

[...]

PR #640, opened 2026-08-25
fix: Allow MultiCOGLayer sampler to be overridden from default `linear`

willjnz:
Related to https://github.com/developmentseed/deck.gl-raster/issues/215#issuecomment-5403656790

Can be simply used like

```tsx
new MultiCOGLayer({
  ...
  bandSampler: { minFilter: "nearest", magFilter: "nearest" },
}),
```

| Before (Linear sampler - blurred, with artifacts) | After (Nearest sampler - sharp, clean) |
|---|---|
| Before 1 | After 1 |
| Before 2 | After 2 |

### DIFF
--- a/packages/deck.gl-geotiff/src/multi-cog-layer.ts
+++ b/packages/deck.gl-geotiff/src/multi-cog-layer.ts
@@ -48,7 +48,7 @@ import {
   metersPerUnit,
   parseWkt,
 } from "@developmentseed/proj";
-import type { Device, TextureFormat } from "@luma.gl/core";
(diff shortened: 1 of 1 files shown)


earlybard:
+1, we need this for `MultiCOGLayer` too. We composite Int16 reflectance and discard nodata with an exact-match filter; with linear sampling every scene edge and nodata hole gets a one-texel rim where the sentinel blends into real data and slips past the filter. Nearest fixes it. It also means `r32float` bands (#650) don't need `OES_texture_float_linear`.

PR #652, opened 2026-09-17
fix(deck.gl-geotiff): raise COG open failures through onError

### LINKED ISSUE #205: Internal errors aren't passed to an `onError` callback
I'd like to give users feedback when a `COGLayer` fails to load, but it appears as though `onError` doesn't get hit for some errors. E.g. I've added this to my layer:

```javascript
      new COGLayer({
        id: "cog-" + geotiffHref,
        geotiff: geotiffHref,
        onError: () => console.log("hi there"),
      })
```

Then, a request that loads the GeoTIFF at https://maxar-opendata.s3.amazonaws.com/events/afghanistan-earthquake22/ard/42/120200201013/2022-06-27/10300100D4928800-visual.tif fails with:

Image

No "hi there". I've dug a bit into **deck.gl** error handling, and haven't come with a solution that doesn't just implement a `this.props.onError?.()` manually. If that's the correct approach, happy to PR that.

earlybard:
## What I am changing

`COGLayer` and `MultiCOGLayer` kick off the GeoTIFF open from `updateState` and never look at the promise again. If the header fetch or CRS lookup fails you get an unhandled rejection in the console, `onError` stays silent, and the layer just sits there empty.

Both now catch that and hand it to `Layer.raiseError`, so `props.onError` fires, and Deck's `onError` after it unless you return `true`.

## How you can test it

```ts
new MultiCOGLayer({
  id: "missing",
  sources: { red: { url: "https://example.com/missing.tif" } },
  composite: { r: "red" },
  onError: (error) => console.warn(error.message, error.cause),
});
```

On `main`: unhandled rejection, no callback. With this: one `loading COG sources: …` call. There's a unit test per layer too, and both fail on `main`.

## Related Issues

Closes #205. Takes over from #206, which did the same for `COGLayer` only and has drifted out of date.

AI disclosure: written with Claude Code as a code assistant, reviewed by me.


kylebarron:
Thanks @earlybard. Can I ask how you're using deck.gl-raster?

earlybard:
Hey @kylebarron. I'm building a COG viewer for open Geoscience Australia data. I'm:

- Streaming 16 bit COGs to the browser, rendering with MapLibre + deck.gl
- Using `MosaicLayer` + `MultiCOGLayer` to render many scenes as a mosaic 
- Rescaling to 8 bit dynamically, allowing histogram manipulation
- Supporting band selection and mixing. 

There are some tricky mosaic-like issues I'll be tackling next, e.g. controlling z-index of scenes within the mosaic, and not rendering fully occluded scenes.

Thanks again!

packages/deck.gl-geotiff/src/cog-layer.ts
```diff
@@ -202,7 +202,9 @@ export class COGLayer<
       // Clear stale state so renderLayers returns null until the new GeoTIFF is
       // ready
       this.clearState();
-      this._parseGeoTIFF();
+      this._parseGeoTIFF().catch((error: Error) =>
+        this.raiseError(error, "loading GeoTIFF"),
```
kylebarron, line 206: Oh interesting is `raiseError` an upstream method in the base deck.gl `Layer`?

earlybard: Yep it is. Marked (Internal) in the docstring but the types are public. Not sure how you feel about that!

My use case is a display in my UI if a GeoTIFF fails to load. 

The alternative is a new prop like onGeoTIFFError, but this seemed easier + hooks into Deck's error handling.

kylebarron (approved):
Thanks!

